PR #1, opened 2015-11-14
Directly load elements instead of via bower

JeremybellEU:
This eases the development of sub-elements.

Previously, updates should be committed and pushed to the respective elements, then a `bower update` in the frontend and then a refresh of the page. This significantly slows the development speed.

The new approach is to stick with the git submodules, but load the elements directly in `app/elements/*` instead of via bower. Therefore if you locally change an element, `gulp serve` automatically refreshes the page.

@svenpopping PTAL


svenpopping:
Helemaal prima!

PR #10, opened 2015-12-24
Add basic login and paging

TimvdLippe:
Implementation of registering and logging in using `lancie-login-check` and `lancie-login-form`. All secured ajax calls should use `lancie-ajax`.

### DIFF
--- a/app/elements/elements.html
+++ b/app/elements/elements.html
@@ -57,7 +57,5 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../bower_components/google-map/google-map-directions.html" />
 <link rel="import" href="../bower_components/google-youtube/google-youtube.html" />
 
-<link rel="import" href="../bower_components/akc-meta/akc-meta.html">
-
 <!-- Configure your routes here -->
 <link rel="import" href="routing.html">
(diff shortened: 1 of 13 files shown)


skamoen:
Even though it's not in master yet, CSRF will be merged soon. Because login is a POST operation, it will also need a valid CSRF token to be sent with the request. I suggest something like this:

When opening the login form, make an AJAX GET request to /login (/token works as well), which will return the login page. You can ignore the response body, but in the header you will find "X-CSRF-TOKEN". You need to save this token, and send it as a parameter to the POST request to /login, as if it was a hidden field.

The response on the login POST will contain a new CSRF token, which you should use for all POST requests for the session, as part of the request header (again as "X-CSRF-TOKEN"). To make sure you always a have a valid token, you can find a valid token in the response of /token. 

See https://github.com/AreaFiftyLAN/lancie-api/blob/fb-master_csrf/src/test/java/ch/wisv/areafiftylan/IntegrationTest.java for the the login() function, which mimics this behaviour

TimvdLippe:
CSRF handling implemented.

TimvdLippe:
@svenpopping could you please also take a look and merge if you also agree?

svenpopping:
:+1:

PR #36, opened 2016-02-15
Ticket page update

Icety:


app/elements/lancie-ticket-page/lancie-ticket-block.html
```diff
@@ -82,10 +83,19 @@ <h3 class="horizontal layout start">Price: € {{price}}</h3>
...
-            return price + ",-";
+            return Math.abs(price) + ",-";
           } else {
-            return price.toFixed(2);
+            return Math.abs(price).toFixed(2);
+          }
+        },
+
+        _ucfirst: function(string) {
+          if (typeof string == 'undefined') {
```
TimvdLippe: This line can be omitted

TimvdLippe:
One small nit. LGTM

Icety:
It is important to check if the string is empty, because charAt might fail if it checks for char 0 on an empty string.

PR #44, opened 2016-02-16
Improve performance by minifying more content

### LINKED ISSUE #42: Investigate usage of Polybuild
To minify the HTML and CSS/JS we could use https://github.com/PolymerLabs/polybuild This is still in the labs, so we have to see if it does not break any existing styling.

TimvdLippe:
Run `npm install` and then `gulp serve:dist` to see the end-result. All files should be minified.

Fixes #42


skamoen:
Doesn't break anything, so LGTM

PR #49, opened 2016-02-18
Fb master frontpage edits

Icety:
Fixed the location of the sponsors images and the ch logo.
Also the information text has been updated.


martijn9612:
Changes look good, but shouldn't the topicus logo be a bit bigger? All the text could be the same size to look a lot cleaner.

app/elements/lancie-sponsors/lancie-sponsors.html
```diff
@@ -27,6 +27,18 @@
         width: 100%;
         max-width: 250px;
       }
+
+      .topicus {
+        margin-top: -33px !important;
```
TimvdLippe: Why are the !important needed here? We should avoid these statements at all cost, as it makes reasoning too hard and creates a maintenance burden.

Icety: I have removed the !important tags.
@martijn9612 Topicus it's logo is way out of proportion because of the lizard, making the text bigger, makes the image way too large, as well as the lizard.

martijn9612:
Looks very good to me, I think we can merge this.

PR #55, opened 2016-02-21
Fix double removal of ticket

### LINKED ISSUE #53: Removing the first ticket removes 2 tickets
Via @Icety https://github.com/AreaFiftyLAN/lancie-api/issues/129 :
When you click the delete button of the first ticket when having 2 or more tickets, both the first and second ticket get visually deleted. 
The first ticket is actually deleted and the second only on the fronted, refreshing the page will show it again.

I discovered that it goes wrong somewhere after the 'ticketReload' method.

martijn9612:
Fixes #53 
The Style of the ticket was removed, causing the code to ignore the ticket and removing the next one in the list.


TimvdLippe:
LGTM

PR #57, opened 2016-02-21
Add Ticket Price To List Of Selected Tickets

martijn9612:
The prices of the individual tickes in your shopping cart are now visible.


Icety:
LGTM.

PR #63, opened 2016-02-23
Allow users to change their passwords through reset

### LINKED ISSUE #58: Request reset password page
If a user forgets its password, an option to request a password reset email needs to be implemented. The API logic is already there, see the `AuthenticationController`. Basically come down to a single input field for the users' email adress. 

Take into account that this shouldn't allow for bruteforcing to check if an email is registered or not!

martijn9612:
Fixes #58


skamoen:
After resetting the password, you should get a message with success, after which you can go to the login screen. There's no feedback on a successful reset.

skamoen:
Looks good otherwise!

TimvdLippe:
LGTM

PR #66, opened 2016-02-23
Fix error for undefined at onMenuSelect

TimvdLippe:


martijn9612:
Looks good to me, this fixes the issue.

PR #67, opened 2016-02-23
Add styled content while the page is loading

TimvdLippe:
This adds temporary content while the page is loading. There is an instant response of the actual page, while all elements are downloaded from the server. Once the client finishes the request and stamps out the content, the replacement is hidden and the actual site is shown.

Run `npm update` to install the dependency `polyclean`.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -78,7 +78,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <div id="replacement-content">
     <div class="header"></div>
     <div class="content">
-      <span>AREA51LAN</span>
+      <span>AREA51AN</span>
     </div>
   </div>
 <template is="dom-bind" id="app">
--- a/app/index.html
+++ b/app/index.html
@@ -50,6 +50,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       width: 100%;
     }
     #replacement-content .content {
+      background-color: #1a2b43;
       text-align: center;
       height: calc(100vh - 64px);
       width: 100%;
--- a/app/index.html
+++ b/app/index.html
@@ -42,6 +42,26 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <!-- Tile icon for Win8 (144x144) -->
   <meta name="msapplication-TileImage" content="images/touch/ms-touch-icon-144x144-precomposed.png">
 
+  <!-- Temporary styling for slow loading pages -->
+  <style>
+    #replacement-content .header {
+      background-color: #1a2b43;
+      height: 64px;
+      width: 100%;
+    }
+    #replacement-content .content {
+      text-align: center;
+      height: calc(100vh - 64px);
+      width: 100%;
+    }
+    #replacement-content .content span {
+      display: block;
+      line-height: calc(100vh - 64px);
+      font-size: calc(100vw * 0.12);
+      color: #ffe574;
+    }
+  </style>
+
   <!-- build:css styles/main.css -->
   <link rel="stylesheet" href="styles/main.css">
   <!-- endbuild-->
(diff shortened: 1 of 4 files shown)


app/index.html
```diff
@@ -52,14 +72,15 @@
...
-  <link rel="import" href="elements/elements.html" async>
-  <!-- endreplace-->
 </head>
-
 <body class="fullbleed layout vertical">
 <span id="browser-sync-binding"></span>
+  <div id="replacement-content">
+    <div class="header"></div>
+    <div class="content">
+      <span>AREA51LAN</span>
```
skamoen: Adten.

gulpfile.js
```diff
@@ -177,6 +178,13 @@ gulp.task('vulcanize', function() {
       inlineCss: true,
       inlineScripts: true
     }))
+    .pipe(polyclean.uglifyJs())
```
TimvdLippe: After some thinking this should probably be just `polybuild()`, have to test it out

app/scripts/app.js
```diff
@@ -9,18 +9,25 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
...
   };
+
+  var link = document.createElement('link');
+  link.setAttribute('rel', 'import');
+  link.setAttribute('href', 'elements/elements.html');
+  link.setAttribute('async', 'true');
+  document.head.appendChild(link);
+
+  app.addEventListener('dom-change', function() {
+    document.querySelector('#replacement-content').style.display = 'none';
```
TimvdLippe: Replace with `getElementById` (http://stackoverflow.com/questions/15046356/performance-using-js-queryselector)

app/scripts/app.js
```diff
@@ -9,18 +9,25 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
...
-        drawerPanel.closeDrawer();
-      }
-    };
+  // Grab a reference to our auto-binding template
+  // and give it some initial binding values
+  // Learn more about auto-binding templates at http://goo.gl/Dx1u2g
+  var app = document.querySelector('#app');
+  // Close drawer after menu item is selected if drawerPanel is narrow
+  app.onMenuSelect = function() {
+    var drawerPanel = document.querySelector('#paperDrawerPanel');
```
TimvdLippe: Replace with `getElementById` (http://stackoverflow.com/questions/15046356/performance-using-js-queryselector)

gulpfile.js
```diff
@@ -290,7 +299,7 @@ gulp.task('default', ['clean'], function(cb) {
     ['copy', 'styles'],
     'elements',
     ['images', 'fonts', 'html'],
-    'build',
+    'vulcanize', 'build',
```
TimvdLippe: `vulcanize` does not reflect the behavior of the task, as it now crushes `elements.html`. Probably should take a look at the naming.

app/index.html
```diff
@@ -42,6 +42,26 @@
   <!-- Tile icon for Win8 (144x144) -->
   <meta name="msapplication-TileImage" content="images/touch/ms-touch-icon-144x144-precomposed.png">
 
+  <!-- Temporary styling for slow loading pages -->
+  <style>
+    #replacement-content .header {
```
TimvdLippe: Maybe fading out replacementcontent and fading in the paperheaderpanel reduces the flash of content

TimvdLippe:
@svenpopping Could you style the loading element?

TimvdLippe:
@skamoen did you have any objections on these changes?

skamoen:
I think the impact on devices that don't need this is too big for the small gain we would have on devices that do. I haven't properly checked this though.

skamoen:
Let's re-evaluate after implementing lazy-loading next year.

PR #78, opened 2016-02-27
Fix undefined error on ticket checkout

### LINKED ISSUE #77: Checkout button: Cannot read property 'generateRequest' of undefined
When clicking the checkout button on the order page, this results in a `Uncaught TypeError: Cannot read property 'generateRequest' of undefined`

TimvdLippe:
Fixes #77


skamoen:
Wauw. How was this even a thing :stuck_out_tongue_closed_eyes:

TimvdLippe:
I wrapped the ticket page in a `dom-if` for the profile check. However, all elements with an id inside a `dom-if` or `dom-repeat` do not populate the `this.$` object. If you extract them outside these elements, they are populated.

PR #81, opened 2016-02-28
Add a tooltip for the sponsor sites

TimvdLippe:
As discovered in #79


skamoen:
Doesn't the sponsor field have a name instead of href?

TimvdLippe:
Only `href` and `logo`: https://github.com/AreaFiftyLAN/lancie-frontend/blob/master/app/elements/lancie-sponsors/lancie-sponsors.html#L56

PR #87, opened 2016-02-28
Change header text as ticket sale is open

### LINKED ISSUE #84: Replace 'Tickets coming soon'
Tickets are coming right now, so replace that text with a link to the ticket sale.

TimvdLippe:
Fixes #84


skamoen:
Would be better if this was an actual link

PR #98, opened 2016-02-29
Use page.js fork for hashbang issue

### LINKED ISSUE #91: Going to a page directly results in 404 first
When going directly to a page (so not through the menu) the page displays a 404. When refreshed, it shows the page anyway. For pages such as the password reset page, this is a problem that needs to be fixed.

TimvdLippe:
Fixes #91

Run `bower install` and `bower update` to fetch the correct versions.


skamoen:
Can't believe this works. Nice job.

PR #108, opened 2016-03-01
Add default text if profile is not yet filled in

### LINKED ISSUE #83: Set default text if profile is not filled in
image

Simply text with: "Not yet filled in" would fill all this empty space.

TimvdLippe:
Fixes #83


skamoen:
While it might be a little more complicated, I think it would be better if this would be a single line somehwere in the center stating "Not yet filled in, please click the [pencil icon] to complete your profile".  I think it's impossible to have a partially filled profile, so this would make sense. Bit harder to style though.

TimvdLippe:
The middle line now contains the message, which is I think informative enough.

skamoen:
Alright, this will probably be redone in another materializing pass of my-area. Much better and good enough for now :+1:

PR #111, opened 2016-03-01
Add message for already claimed token

### LINKED ISSUE #105: Catch invalid/used verification token response
The verification page now has a single error page. This can be expanded to include an expired / already used token message.

TimvdLippe:
Fixes #105


app/elements/lancie-login/lancie-register-confirmation.html
```diff
@@ -15,9 +15,30 @@
...
+    <lancie-message-card hidden$="{{!error}}">
+      Hmm, there seemed to have gone something wrong with the verification of your account.
+      Please <a href="mailto:lancie@ch.tudelft.nl">send us an email</a> containing your email-adres and username.
+      We will then check what went wrong.
+
+      <a href="/">Head back to the homepage</a>
+    </lancie-message-card>
+
+    <lancie-message-card hidden$="{{!tokenTwice}}">
+      Hmm, this token was already claimed. If you were not the one who claimed the token, please <a href="mailto:lancie@ch.tudelft.nl">send us an email</a> containing your email-adres and username.
```
skamoen: It's _extremely_ unlikely that a token was claimed by someone else. Better say something "Did you already confirm your account?"

TimvdLippe: Fixed :+1:

skamoen:
Now that I think of it, the verification token expires in 24 hours... Once you miss that window, there's no way of requesting a new token. Probably something I should look into in the API. 

Let's keep this open until I figure out how to deal with expired tokens.

skamoen:
As per discussion on slack and https://github.com/AreaFiftyLAN/lancie-api/issues/152. The API will be updated to remove users with expired tokens. Therefore, the expired token will simply result in a 404, which is handled. This PR is therefore sufficient for that usecase. We'll deal with cases where the token expires manually until then.

PR #112, opened 2016-03-01
Change gender field width

### LINKED ISSUE #72: Profile gender field layout
The profile gender field should be aligned better.

martijn9612:
Fixes #72 
Gender field width is decreased by increasing the width of the `displayName` field.


TimvdLippe:
LGTM

PR #117, opened 2016-03-05
Change deprecated shadow flex-layout to new files

TimvdLippe:
The old files are deprecated and generate a lot of warnings. Switch to the new standard.


TimvdLippe:
Reference https://github.com/PolymerElements/iron-flex-layout/pull/32

TimvdLippe:
Requires `bower update` to fetch the new file of `iron-flex-layout`

skamoen:
Bower update did the trick, LGTM

PR #120, opened 2016-03-06
Order return page

### LINKED ISSUE #116: Ticket order success callback
If you cancel an order during payment, you still get redirected to the success page. The ticket also shows up in my-area. This maybe needs to be made more explicit to let the user know that the ticket was actually not ordered. We got this email mentioning a bug, although this is not a bug, it is an issue.

image

### LINKED ISSUE #118: Order succes page is accessible when logged out
Should only be available when actually buying a ticket and the process being a succes.
https://areafiftylan.nl/#!/ordersuccess

skamoen:
When coming back from the Mollie environment, the page fetches the latest status from Mollie, actually stating what the status is, instead of assuming success.

Depends on https://github.com/AreaFiftyLAN/lancie-api/pull/154

Fixes #118 
Fixes #116


app/elements/lancie-order-check.html
```diff
@@ -0,0 +1,69 @@
...
+      this.success = false;
+      this.error = false;
+      this.cancelled = false
+      this.waiting = false;
+      this.orderId = window.location.href.split('order=')[1];
+      this.$.ajaxConfirmOrder.generateRequest();
+    },
+
+    onResponse: function(e, request) {
+      // Such beauty Tour-De-France-code
```
TimvdLippe: :laughing: :+1:

TimvdLippe:
Maybe add a interim `lancie-message-card` (just like the registration) that says "We are confirming your ticket, one sec..."

TimvdLippe:
I assume it is working. Code LGTM.

PR #130, opened 2016-03-07
Enable caching on Travis

TimvdLippe:


martijn9612:
Looks good to me, chaching works and builds are significantly faster.

PR #131, opened 2016-03-07
Hotfix incorrect colors for forms

TimvdLippe:
This is actually an upstream bug which should be fixed in Polymer.
Reproducible with gulp `serve:dist` with minified CSS and a
mix-in without brackets


martijn9612:
Looks good to me, works as intended.

PR #152, opened 2016-03-13
Enable shadow dom and fix minor styling issues

### LINKED ISSUE #128: Consider enabling shadow-dom
Shadow dom is an performance increase for the browser that support it. The old browser will continue using all polyfills.

We need to investigate what CSS and JS is breaking. This is most likely only an issue in `index.html` with its files `app.js` and `main.css`.

TimvdLippe:
There were some styling issues with the activity dialogs, they should be properly using mix-ins now. Also `:host(:after)` appears to be illegal in shadow dom, therefore these rules are moved to `lancie-sponsors`.

Fixes #128

### DIFF
--- a/app/elements/lancie-activity/lancie-activity-dialog.html
+++ b/app/elements/lancie-activity/lancie-activity-dialog.html
@@ -23,6 +23,9 @@ This code may only be used under the BSD style license found at https://github.c
 <dom-module id="lancie-activity-dialog">
   <style include="iron-flex iron-flex-alignment"></style>
   <style>
+    :host {
+      display: block;
+    }
     paper-dialog.small-dialog {
       position: fixed;
       top: 0;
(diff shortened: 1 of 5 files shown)


TimvdLippe:
Shadow dom is only used if it is: 1. Available 2. Enabled. Therefore this PR will only result in using shadow dom on Chrome (currently). For all other browsers, the current shady dom is used and will not see any (performance) difference.

skamoen:
It looks to me like it's less smooth on chrome then Shady DOM. Animations look smoother and on loading I can see some flickering. Hard to reproduce or show, but it doesn't look like an improvement...

TimvdLippe:
I am unable to reproduce your flickering. Are you sure you have all up-to-date versions? `npm update && bower update`. Else I need to see it on your laptop to investigate the issue.

TimvdLippe:
Somehow a lot of errors appeared in the console. Mostly because all `<content>` in `lancie-login-check` is stamped, whereas it shouldn't. So far I have not been able to track down the root cause.

TimvdLippe:
Blocked by the resolution of https://github.com/Polymer/polymer/issues/3521

skamoen:
Still blocked. This is something for next year. See #128

PR #157, opened 2016-03-16
Clean up gulpfile and apply versioning to index.js

### LINKED ISSUE #154: Use version number in .js and .html to enable caching
Currently we cache the .js and .html 24 hours. It would be better if we use a version number in the file name and cache it indefinitely. Then we push a new version live, the name of the file is different and the updated file will be downloaded.

This most likely requires configuration in the `gulpfile.js` and we must see if a file like `index.v121.html` is able to be served.

TimvdLippe:
`index.html` can not be versioned sadly, but we can version `index.js`. This allows infinite caching of `.js` since it will automatically redownload if the hash of the `.js` is changed. Also I cleaned up a lot of old tasks that are already handled by `polybuild` in the `build` step.

Run `npm install` to download the new dependencies.

Fixes #154

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -28,34 +28,10 @@ var polybuild = require('polybuild');
 var jsonminify = require('gulp-jsonminify');
 var jshint = require('gulp-jshint');
 
-var AUTOPREFIXER_BROWSERS = [
-  'ie >= 10',
-  'ie_mob >= 10',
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Ping @skamoen would like to land this asap to reduce amount of rebases necessary for Travis

skamoen:
LGTM

PR #160, opened 2016-03-18
Bump Polymer to 1.4.0 and fix some styling issues

TimvdLippe:
The styling issue in 1.3.0 has been fixed in 1.3.1, so we can once again use the latest version of Polymer.

### DIFF
--- a/app/elements/lancie-activity/lancie-activities.html
+++ b/app/elements/lancie-activity/lancie-activities.html
@@ -13,7 +13,7 @@ This code may only be used under the BSD style license found at https://github.c
 
 <dom-module id="lancie-activities">
   <template>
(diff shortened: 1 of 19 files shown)


skamoen:
Can't we make this just a fixed version, instead of the `^` thing? For example right now, I can only choose between 1.2.4 and 1.4.0. Less choice is better when it comes to building.

TimvdLippe:
Up for review again. After discussion on Slack, we decided to stick with the `^` for now. The 1.2.4 resolution was due to an outdated dangling dependency.

The `<style include="">` should actually be closed, as it broke the parsing of CSS. Including this in the PR.

skamoen:
lgtm

PR #182, opened 2016-04-24
Ticket transfer functionality in my area

### LINKED ISSUE #142: Ticket transfer in My Area
Ticket transfers should be done from the Orders & Tickets section in My Area. Let's come up with what you should be able to do that makes sense.

- [x] The User has an overview of its Orders
- [x] The User must be able to see which Tickets were part of that Order, by clicking on the ( i ) button
- [ ] Order Details dialog should contain everything there is to know about that Order, and the Tickets that it included. This means current owner, transfers made (what more?)
- [x] It should list all Tickets the User currently _owns_
- [x] It should not list Tickets that have been transferred, that belongs in the Order details dialog
- [x] For each Ticket that the User owns, there should be a button to transfer that Ticket. Should this be based on username or email?
- [ ] If a transfer is sent, the icon should change to reflect the open transfer (a button to a Transfer details dialog?)
- [ ] The User needs a way to cancel the transfer. (from the Transfer details dialog?)
- [ ] Initiating a Transfer should state some terms, like CH-member and Pickup Service validity

Open questions:

1. Should the User be able to transfer _all_ of his Tickets, or should one remain locked?

martijn9612:
Partially fixed #142 
The user is now able to set up a ticket transfer and cancel ticket transfers.

### DIFF
--- a/app/elements/lancie-my-area/my-area-ticket-item.html
+++ b/app/elements/lancie-my-area/my-area-ticket-item.html
@@ -1,0 +1,65 @@
+<link rel="import" href="../../bower_components/paper-item/paper-item.html">
+<link rel="import" href="../../bower_components/paper-item/paper-item-body.html">
+<link rel="import" href="../../bower_components/paper-dialog/paper-dialog.html">
+<link rel="import" href="../../bower_components/paper-dialog-scrollable/paper-dialog-scrollable.html">
+<link rel="import" href="../../bower_components/paper-icon-button/paper-icon-button.html">
+<link rel="import" href="../../bower_components/iron-icon/iron-icon.html">
+<link rel="import" href="../../bower_components/iron-flex-layout/iron-flex-layout-classes.html">
+<link rel="import" href="../../bower_components/neon-animation/neon-animations.html">
+
+<dom-module id="my-area-ticket-item">
+  <template>
+    <style include="iron-flex iron-flex-alignment"></style>
+    <style>
+    :host {
+      display: block;
+    }
+
+    hr {
+      display: block;
+      height: 1px;
+      border: 0;
+      border-top: 1px solid #e8e8e8;
+      padding: 0;
+    }
+    </style>
+
+    <hr>
+    <paper-item>
+      <paper-item-body class="layout horizontal center justified wrap">
+        <div><b>Type: </b>{{ticket.text}}</div>
+        <div><b>Amount: </b>&euro; {{ticket.price}}</div>
+        <div>CH member:
+          <iron-icon icon="{{getIcon(ticket.chMember)}}"></iron-icon>
+        </div>
+        <div>Pickup service:
+          <iron-icon icon="{{getIcon(ticket.pickupService)}}"></iron-icon>
+        </div>
+        <div>
+          <paper-icon-button icon="icons:send" on-tap="ticketTransfer" disabled></paper-icon-button>
+          <paper-tooltip offset="0" animation-delay="1">Transfer this ticket</paper-tooltip>
+        </div>
+      </paper-item-body>
+    </paper-item>
+  </template>
+  <script>
+    'use strict';
+
+    Polymer({
+      is: 'my-area-ticket-item',
+
+      properties: {
+        ticket: {
+          type: Object,
+          notify: true
+        }
+      },
+      getIcon: function(binding) {
+        return binding ? 'icons:check' : 'icons:clear';
+      },
+      ticketTransfer: function() {
+        this.$.ticketTransfer.toggle();
+      }
+    });
+  </script>
+</dom-module>
--- a/app/elements/lancie-my-area/my-area-ticket-item.html
+++ b/app/elements/lancie-my-area/my-area-ticket-item.html
@@ -22,9 +22,14 @@
       border-top: 1px solid #e8e8e8;
       padding: 0;
     }
+
+    .cancelTicketTransfer {
+      color: red;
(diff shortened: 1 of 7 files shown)


skamoen:
Alignment on the different ticket types is a bit off.
image

TimvdLippe:
image

skamoen:
Following the link from the email results in a 404. 

"Jan has send you a ticket for AreaFiftyLAN! To accept this ticket please click on the following link: https://areafiftylan.nl/#!/acceptTransfer/?token=e0572a9f-0e3c-4db7-b7dd-1c6c3ed6d4a8"

Oops, you hit a 404. Head back to the homepage.

app/elements/lancie-my-area/my-area-ticket-item.html
```diff
@@ -0,0 +1,163 @@
...
+        <div><b>Type: </b>{{ticket.text}}</div>
+        <div><b>Amount: </b>&euro; {{ticket.price}}</div>
+        <div>CH member:
+          <iron-icon icon="{{getIcon(ticket.chMember)}}"></iron-icon>
+        </div>
+        <div>Pickup service:
+          <iron-icon icon="{{getIcon(ticket.pickupService)}}"></iron-icon>
+        </div>
+        <div>
+          <paper-icon-button icon="icons:send" on-tap="openTicketTransferDialog"></paper-icon-button>
```
skamoen: The icon should change to something else then the arrow if a click will result in a cancel action

skamoen:
input fields aren't cleared after a succesful transfer

image

martijn9612:
I am fairly sure that the last one is only an issue due to the way polymer behaves when using 2 accounts on the same pc in different browser windows because of the special developer behavior. Were you currently using 2 browser windows at the same time?

app/elements/lancie-my-area/my-area-ticket-item.html
martijn9612: For because the request is sent after the page is rendered, the `dom-if` does not work on the `{{ticket.token}}` property. At the time the page is rendered, this variable is still undefined. This results in the `dom-if` not working. I would love to find out a way to get this working.

TimvdLippe: Sorry but I do not understand what is going wrong. I do not see a `dom-if`. Could you elaborate on what you were trying to do?

martijn9612: I did indeed remove the `dom-if`. What I am currently doing is requesting the tokens and tickets seperately and joining the tickets with the tokens. I want to check whether the ticket contains a token or not with a `dom-if`. The token is not correctly read or set later, thus the `dom-if` doesn't function at all and just removes both cases from the page, resulting in no send button whatsoever.

skamoen:
Hello data leak:

```JSON
{  
   "status":"200",
   "timestamp":"2016-04-27T11:29:44.669",
   "message":"Ticket successfully set up for transfer.",
   "object":{  
      "token":"926ab73a-1fae-4d50-9864-814eae10257a",
      "user":{  
         "username":"user2",
         "email":"bert@mail.com",
         "profile":{  
            "firstName":"Bert",
            "lastName":"Kleijn",
            "displayName":"ILoveZombies",
            "gender":"OTHER",
            "address":"Mekelweg 20",
            "zipcode":"2826CD",
            "city":"Amsterdam",
            "phoneNumber":"0611",
            "notes":null
         },
         "id":3,
         "reference":111578567,
         "authorities":[  
            "ROLE_USER"
         ]
      },
      "expirable":false,
      "expiryDate":[  
         2016,
         4,
         27,
         11,
         29,
         44,
         188000000
      ],
      "ticket":{  
         "id":11,
         "owner":{  
            "username":"user",
            "email":"user@mail.com",
            "profile":{  
               "firstName":"Jan",
               "lastName":"de Groot",
               "displayName":"MonsterKiller9001",
               "gender":"MALE",
               "address":"Mekelweg 4",
               "zipcode":"2826CD",
               "city":"Delft",
               "phoneNumber":"0906-0666",
               "notes":null
            },
            "id":1,
            "reference":3599307,
            "authorities":[  
               "ROLE_USER",
               "ROLE_ADMIN"
            ]
         },
         "type":"EARLY_FULL",
         "text":"Early Bird",
         "pickupService":false,
         "chMember":false,
         "valid":true,
         "price":37.5
      },
      "valid":true
   }
}
```

martijn9612:
The data leak is not an issue in the frontend, I created an issue in the backend for this.

TimvdLippe:
* The notification that the transfer is successful is too small and in the bottom left corner. Preferably this should be more obvious
* The button to transfer is not obvious. If you don't know you can transfer, you will likely not click this arrow button. Probably add `Transfer:` to it
* The e-mail says `Click`, but you can't click the link
* I get a "You are not logged in message" when opening the link, but this is removed after ~2 seconds
* The terms and conditions stayed very small upon opening, but then increased in size.
* When I get linked back to my-area, the ticket is not there.
* While waiting for the server to verify the transfer (after accepting the T&C), set a message "Accepting ticket..." to indicate it is loading

martijn9612:
As for the `Transfer:` text and the size of the successful toast, we should maybe talk about this. Since these are not big issues. We could fix this later and add an issue in git. The `Transfer:` text would be easy to fix, the toast would involve quite some work designing it since there is not an easy alternative.

svenpopping:
LGTM

PR #183, opened 2016-05-04
Updated Master Tournament Info

bvanderdrift:
## Removed Melee
We're still in discussion with DSEA to make this official

## Altered LoL description
To be able to be recognized as an official LoL tournament by RIOT, we need to comply with some rules. We have been rejected with the following feedback regarding our website:

> What information do I need to have available when submitting my event application? You’ll need to have your rules, competition structure (bracket or round robin, etc.) and tournament dates ready and publicly visible on your event homepage. If you don’t include your tournament rules with your submission and post your rules publicly, for example, your event will be rejected. Which mode will be the tournament played at, is it best of 1/3/5, how many teams allowed, how many participants per team allowed?

Could you also look into if you think this feedback is resolved with this PR?

### DIFF
--- a/app/scripts/activity.json
+++ b/app/scripts/activity.json
@@ -39,7 +39,7 @@
     "title": "LoL",
     "subtitle": "5 v 5",
     "headerTitle": "League of Legends",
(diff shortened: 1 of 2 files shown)


skamoen:
LGTM

PR #191, opened 2016-05-07
Improve layout navigation menu

svenpopping:


skamoen:
Yes, more fancy is more better.

PR #201, opened 2016-05-12
Discord Widget

skamoen:
This puts a Discord card in My Area. It uses the Discord API to show the current number of online people and provides a link for easy access. 

capture2


martijn9612:
Looks good, but is there no way to directly link to the server? I would like to click the join button and just go to the server directly. Is this possible?

skamoen:
This is an invite for users that aren't already in there. New users need this to get acccess. I think a second button directly to discordapp would be overkill.

martijn9612:
Makes sense, although you can still join with this link, it is just one click more. Looks good, merging!

PR #210, opened 2016-05-19
Seatmap implementation

martijn9612:


martijn9612:
Unfortunately we cannot remove the reservationError variable, when using the reservationErrorMessage only the result is that the red error field is always showing.

skamoen:
Run `gulp jshint` to find out where your build is failing

PR #217, opened 2016-05-20
Fix route tabs

martijn9612:
Additionally fix the default opened tab on the timetable.

### DIFF
--- a/app/elements/lancie-header/lancie-header.html
+++ b/app/elements/lancie-header/lancie-header.html
@@ -32,7 +32,7 @@
     }
     </style>
 
-    <paper-tabs selected="0">
+    <paper-tabs selected="{{_selected()}}">
       <paper-tab link>
         <a href="{{baseUrl}}/" tabindex="-1">Home</a>
       </paper-tab>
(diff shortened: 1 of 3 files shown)


app/elements/lancie-header/lancie-header.html
```diff
@@ -56,6 +56,18 @@
         type: String,
         notify: true
       }
+    },
+    _selected: function() {
+      var route = window.location.href.split('#!/')[1];
```
TimvdLippe: Carbon route would fix this btw. But I think it is unwise to switch as this stage, since it has a major effect on your architecture. Just an FYI

TimvdLippe:
Glad I did not merge yet and tried to run locally. This does not work if you do not have a hash set. E.g. you open `localhost:5100/` (or `areafiftylan.nl` for that matter) then route is undefined. Probably the cleanest solution is to add `if (!route) return 0;`

martijn9612:
I have found a solution but the route property never contains anything. The variable is not used, so I guess it can be removed.

martijn9612:
This commit can be squased! This can be done by selecting the arrow on the right when you click the merge button.

Icety:
LGTM, but cant test this in a browser right now

PR #219, opened 2016-05-21
404 page styling

martijn9612:


Icety:
Much better than the previous 404!
LGTM

PR #220, opened 2016-05-22
Add message when the csrf is expired

martijn9612:


Icety:
Shouldnt this also be done at the pages that rely on csrf?
Besides that LGTM.

martijn9612:
The rest of the pages are behind a login, the login has the refresh message.

PR #226, opened 2016-05-22
listbox was not visible in the dom-if

martijn9612:
@TimvdLippe Could you look at this? If the listbox is in the dom-if the `this.$.listbox` selector returns undefined. I have no clue why.


TimvdLippe:
https://www.polymer-project.org/1.0/docs/devguide/local-dom#node-finding

> Note: Nodes created dynamically using data binding (including those in dom-repeat and dom-if templates) are not added to the this.$ hash. The hash includes only statically created local DOM nodes (that is, the nodes defined in the element's outermost template).

You should use  `hidden$="{{!tickets.length}}"` on a surrounding div instead of that `dom-if`

PR #232, opened 2016-05-25
Add faq section to the homepage

martijn9612:
@Exclaminator Still needs to add frequently asked questions to this section.

PR #233, opened 2016-05-25
Add terms of service button to my-area

martijn9612:


skamoen:
LGTM. Should also be in the FAQ (#232)

PR #237, opened 2016-05-27
Added a parameter for force-updating cached json

bvanderdrift:


skamoen:
Maybe add sponsors as well?

martijn9612:
Everything still works, wouldn't have expected otherwise.

skamoen:
Not sure how to test this, but at least the JSON's work fine here, so I assume browser-magic will fix the cache stuff

martijn9612:
I believe this does not apply to the sponsors. They are on their own element, and do not use json.

skamoen:
Good point, I see they use inline json which doesn't have the issue indeed.

PR #239, opened 2016-05-28
Change the check to only apply to early bird tickets

martijn9612:


bvanderdrift:
Please next time refer to an issue before sending in a PR.

After internal discussion with @martijn9612 we noticed this was not the correct way to fix the problem. The problem was that this check passed and made the tickets unavailable in production. This fixed itself, so the problem needs to be investigated further and a correct fix needs to be made.

PR #240, opened 2016-05-29
New banner message

martijn9612:


skamoen:
I'd put something with last minute here as well, to stress people are already late to the party.

"Last minute tickets available until the 2nd of June!"

bvanderdrift:
LGTM

martijn9612:
Good one!

PR #252, opened 2016-05-31
Add min width

### LINKED ISSUE #223: Lower row in seat groups on the left have wrong widths.
image

martijn9612:
Fixes #223 
For some reason a min-width does not require a value to fix the spacing in the lower row.


svenpopping:
LGTMBNIY

PR #261, opened 2016-06-01
Feature rfid

martijn9612:
* Checking in users
* Checking whether a user is allowed to drink alcohol
* Consumption tracking


svenpopping:
LGTM

PR #263, opened 2016-06-03
Fixed seatmap

svenpopping:


app/elements/lancie-seatmap/lancie-seat.html
```diff
@@ -132,39 +142,36 @@
         }
       }
     },
-    _calculateClass: function(user, teamTickets, seat) {
+    _calculateClass: function(user, teamTickets, seat, filter) {
+      console.log(seat);
```
martijn9612: Should be removed.

martijn9612:
Looks good apart from a minor comment!

PR #269, opened 2016-06-06
Updated Banner text

bvanderdrift:
So the event is over, let's update the banner on the frontpage!


skamoen:
I think more needs to be updated then just the banner text :-P

martijn9612:
Smart usage of the text. LGTM.

PR #277, opened 2016-09-02
Fix min page height & remove unneeded css properties

mgajanssen:
The page no longer scrolls when the content is not long enough. :tada:


TimvdLippe:
LGTM

PR #281, opened 2016-10-14
Change login method & email check

### LINKED ISSUE #279: Fix registration form
As of now, the registration for is incorrect. The user can enter two unequal passwords and still have a successful account creation when a correct email address is entered. This issue is caused by this part of the code in `lancie-login-form.html`:

``` javascript
validForm: function() {
    var valid = true;
    valid = this.validPassword();
    valid = this.validEmail();
    return valid;
},
```

The code does not care about a false or incorrect password, this needs to be resolved.

### LINKED ISSUE #280: Change registration form email field
Currently the email validation in the account creation window is custom. We can switch this out for a `gold-email-input`, this does everything automatically and reduces logic in `lancie-login-form.html`. This also has a nice way of displaying error messages.

martijnjanssen:
Fixes #279 #280

### DIFF
--- a/src/lancie-login/lancie-login-form.html
+++ b/src/lancie-login/lancie-login-form.html
@@ -14,6 +14,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../../bower_components/paper-card/paper-card.html">
 <link rel="import" href="../../bower_components/paper-material/paper-material.html">
 <link rel="import" href="../../bower_components/paper-input/paper-input.html">
+<link rel="import" href="../../bower_components/gold-email-input/gold-email-input.html">
 <link rel="import" href="../../bower_components/paper-button/paper-button.html">
 
 <dom-module id="lancie-login-form">
@@ -89,7 +90,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       url="/api/v1/users"
       method="POST"
       handle-as="json"
-      body="{{bodyFrom(username, password, email)}}"
+      body="{{bodyFrom(username, password)}}"
       headers="{{headerFrom('application/json', csrfToken)}}"
       on-response="onRegister"
       on-error="onRegisterError"></iron-ajax>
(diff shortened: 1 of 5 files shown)


app/elements/lancie-login/lancie-login-form.html
```diff
@@ -105,12 +106,10 @@
         </div>
```
TimvdLippe: The input of the login form should be a gold-input too

app/elements/lancie-login/lancie-login-form.html
```diff
@@ -203,7 +203,12 @@
       },
 
       tryLogIn: function() {
-        this.$.ajaxLogin.generateRequest();
+        if (this.$.username.validate()) {
+          this.$.ajaxLogin.generateRequest();
+        } else {
+          this.loginError = false;
+          this.loginErrorMessage = '';
```
TimvdLippe: Is this intended? Also we should check for the password field filled in I suppose?

mgajanssen: Yes this is intended, there is no need to check the password, upon registering the password has to have a length. The password cannot be empty, the server returns `Bad credentials`, this is also shown in the login screen.

mgajanssen: The password check is performed in the [lancie-api](https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/users/model/UserDTO.java#L28-L31).

src/lancie-login/lancie-login-form.html
```diff
@@ -279,39 +276,31 @@
...
+          this.validatePassword();
           return false;
         }
-        this.equalPassword = true;
-        return true;
+        return this.validatePassword();
+      },
+
+      validateUsername: function() {
+        return this.$.registerUsername.validate();
```
TimvdLippe: This does not validate the username email input when logging in. While the two inputs have the same data-binding, the error is only shown on the registering email input.
It should therefore invoke both validate functions, but the latter result is used since the databinding is equivalent.

mgajanssen: This function has nothing to do with validation of the login, this method is only called in the `validateForm()` method, which is only called in the `tryRegister()` method. The function name needs to be changed though.

TimvdLippe: I would expect that if I enter an invalid email, lets say `foo`, into the login input that I get a nice message saying it is not an email.

mgajanssen: That is exactly what the `validate()` method of the `gold-email-input` does. It shows a message that the text put in is not a valid email.

TimvdLippe:
I think the forms need to be validated `on-keydown`. Other than that it is now working on my machine :tada:

martijnjanssen:
I think all issues are fixed now. Back to you!

PR #285, opened 2016-10-25
Polymer build

TimvdLippe:
This updates the project file structure for Polymer build. You can generate a build with `npm run build`. You can inspect the build with `npm run serve-build`. For serving your local dev project you run `npm run serve`.

PR #286, opened 2016-10-26
Fix proxy address to correct api location

mgajanssen:


TimvdLippe:
Yup, oopsie forgotten in #285

Issue #283, opened 2016-10-21
Rework client architecture to be compatible with polymer-build


TimvdLippe:
The current client architecture is incompatible with Polymer build. To be able to generate a PRPL-ready application, several changes are required:
- [ ] Rework `lancie-content` to allow lazy-loading of pages for faster first-paint
  - Adopt `iron-lazy-pages` to lazy-load pages
  - Check all imports to be correctly resolved
- [ ] Adopt a service-worker for faster cache results for static files (I should have a compatible one lying around somewhere)
- [x] Move files in `app/` to `src/` or `/`. All files should be served from `/` instead of `app/`, which means that files like `index.html` and `manifest.json` need to be in root. However, all other elements can remain in `src/` (which is a simple remain to be consistent with all other frontend projects)
- [x] Replace `page.js` with `app-route`

This is all based on https://github.com/PolymerElements/generator-polymer-init-custom-build and similar efforts I already did on other projects.

TimvdLippe:
We will also need to adopt app-route to be able to efficiently load data in sub-routes.

TimvdLippe:
A PR is ready for the second point at #285

martijnjanssen:
The third part is done right now at #290

martijnjanssen:
How relevant is this issue: #225 ? Looks almost identical to me, am I missing something?

skamoen:
It's the same thing.

TimvdLippe:
Only todo is SW, but we can do that later. Main architecture has been updated so the new LANcie can work on features :tada:

PR #287, opened 2016-10-26
Replace page.js with app-route

TimvdLippe:
Another step for #283 `page.js` has been replaced with `app-route`.

Note: this PR removes the hash-routing and rewrites some routes. We might want to take another good look and update the server accordingly.


src/lancie-content.html
```diff
@@ -99,7 +105,7 @@
         <lancie-request-password-reset-success></lancie-request-password-reset-success>
       </div>
 
-      <div data-route="order-check">
+      <div data-route="ordersuccess">
```
mgajanssen: I think a better rename would be `order-success`, this is in line with the other pages regarding transferring/ordering.

src/lancie-layout/lancie-layout.html
```diff
@@ -69,11 +69,11 @@
...
         <paper-menu selected="_selected">
-            <div><a href="{{baseUrl}}/"><paper-item on-tap="_closeDrawer">Home</paper-item></a></div>
-            <div><a href="{{baseUrl}}/tickets"><paper-item on-tap="_closeDrawer">Tickets</paper-item></a></div>
-            <div><a href="{{baseUrl}}/my-area"><paper-item on-tap="_closeDrawer">My Area</paper-item></a></div>
-            <div><a href="{{baseUrl}}/seatmap"><paper-item on-tap="_closeDrawer">Seatmap</paper-item></a></div>
-            <paper-item on-tap="toShop">Shop</paper-item>
+          <div><a href$="{{route.prefix}}/../"><paper-item on-tap="_closeDrawer">Home</paper-item></a></div>
+          <div><a href$="{{route.prefix}}/tickets"><paper-item on-tap="_closeDrawer">Tickets</paper-item></a></div>
+          <div><a href$="{{route.prefix}}/my-area"><paper-item on-tap="_closeDrawer">My Area</paper-item></a></div>
+          <div><a href$="{{route.prefix}}/seatmap"><paper-item on-tap="_closeDrawer">Seatmap</paper-item></a></div>
```
mgajanssen: All the `href` links in here can be replaced to just an identifier corresponding to the items in lancie-content `page` values.

mgajanssen:
Okay, this should do the job quite nicely. Back to you Tim!

TimvdLippe:
Removing `<a>` removes the ability to `ctrl + click` and is less nice for crawlers of search engines. We need to find a solution that makes use of `<a>` but is not crap :smile:

src/lancie-app.html
```diff
@@ -10,15 +10,16 @@
         display: block;
       }
     </style>
-    <lancie-layout route="{{route}}">
+    <lancie-layout page="{{page}}">
       <div id="content">
-        <lancie-content route="{{route}}"></lancie-content>
+        <lancie-content page="{{page}}"></lancie-content>
```
TimvdLippe: Let's name it route.

TimvdLippe:
Yes is working now, nice!

PR #289, opened 2016-10-29
Remove sourceGlobs values that exist by default

MatthijsKok:
as seen in [Link](https://github.com/Polymer/polymer-cli/blob/b5f5f59a348362e612169fcd9efd8982d1076b8e/src/project-config.ts#L17), these values are assumed and do not need to be included. Therefore we can remove them safely.


TimvdLippe:
Ah good catch, TIL.

PR #303, opened 2016-11-11
Fix missing validation method for login username

TimvdLippe:
Was previously throwing warnings in the console when you were typing in the username field of the login form.


martijnjanssen:
Good catch, thanks for fixing this! I also saw something else we missed. :disappointed:

PR #311, opened 2016-11-23
Update website info to AreaFiftyLAN 2017

MatthijsKok:
Location, prizes, tournaments and other info updated.


src/lancie-home-page/lancie-home-page.html
```diff
@@ -86,19 +88,21 @@
...
           <br /><sub>*CH members/alumni will get a small discount</sub>
         </p>
+        <!--
         <p>
           Besides that you can play any game you like, there are organized activities during the weekend.
           Below you can find a global overview. Please note this is subject to change and of course this is not all!
           A final schedule will be provided during the event.
         </p>
         <lancie-timeline></lancie-timeline>
+        -->
```
martijnjanssen: You can just remove the entire timeline, we can update it and put it back in another time.

MatthijsKok:
I agree these are incomplete, however we want to push this new information as fast as possible.

TimvdLippe:
Then the other sections have to be updated too asap. We do not merge incomplete PRs

martijnjanssen:
The entire homepage has to be updated, it is not needed to fill in all the information as of now. We can leave parts of them unfilled and not visible. But the parts that are staying should be updated and adapted for the coming event.

PR #312, opened 2016-11-24
WIP auth token

### LINKED ISSUE #310: New login method for authentication
The new login methods differ quite a bit from the old methods, first of all, there is now a method to authenticate without the use of csrf. The new method of authentication is fairly straightforward:

Firstly, do a `POST` request on `/api/v1/token` with the body containing the `'username'` and `'password'`. By doing so, you get a response with the newly acquired token in the `object`. After this, every request can be authenticated with the header parameter `X-Auth-Token`, granting access to the user. A token is valid for 5 days.

A logout functionality will be implemented in the future, this can create the possibility the token is expired even faster. We should take this into consideration with the new request methods.

TimvdLippe:
Have to go, but this is what I got so far. Probably a lot is still not working.

Fixes #310

### DIFF
--- a/src/lancie-content.html
+++ b/src/lancie-content.html
@@ -23,6 +23,11 @@
       handle-as="json"
       on-response="storeUser"></iron-ajax>
 
+    <iron-ajax
+      url="/api/v1/token"
+      handle-as="json"
+      ></iron-ajax>
+
     <lancie-ajax id="ajax"></lancie-ajax>
 
     <app-location route="{{route}}"></app-location>
@@ -128,8 +133,9 @@
         }
       },
       storeUser: function(event, request) {
-        this.user = (request.response.message === 'Not logged in') ? '' : request.response;
-        this.$.ajax.injectToken(request.xhr.getResponseHeader('X-CSRF-TOKEN'));
+        if (request.response.message !== 'Not logged in') {
+          this.user = request.response;
+        }
       },
 
       _matches: function(path, route) {
--- a/src/lancie-content.html
+++ b/src/lancie-content.html
@@ -17,7 +17,7 @@
       }
     </style>
 
-    <lancie-auth-storage id="auth" user={{user}}></lancie-auth-storage>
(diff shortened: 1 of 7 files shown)


TimvdLippe:
@martijnjanssen Can you take this one?

src/lancie-content.html
```diff
@@ -17,13 +17,7 @@
...
     </style>
 
-    <iron-ajax
-      auto
-      url="/api/v1/users/current"
-      handle-as="json"
-      on-response="storeUser"></iron-ajax>
-
-    <lancie-ajax id="ajax"></lancie-ajax>
+    <lancie-auth-storage id="auth" user={{user}}></lancie-auth-storage>
```
TimvdLippe: Forgot quotes around {{user}}

src/lancie-storage/lancie-auth-storage.html
```diff
@@ -0,0 +1,85 @@
...
+        // Check if token has expired
+        if (!this.auth || this.auth.created + 5 * 24 * 60 * 60 * 1000 < Date.now()) {
+          return;
+        }
+        return this.auth.token;
+      },
+
+      storeToken: function(token) {
+        this.set('auth', {
+          created: Date.now(),
```
TimvdLippe: Isnt this created on the server?

src/lancie-storage/lancie-auth-storage.html
```diff
@@ -0,0 +1,85 @@
...
+          token: token,
+        });
+      },
+
+      storeUser: function(e, request) {
+        if (request.succeeded) {
```
TimvdLippe: Show an error if it failed

src/lancie-storage/lancie-auth-storage.html
martijnjanssen: Oh yeah, you are correct, I was planning on changing that later but I forgot about it.

TimvdLippe:
Branch is now working :tada:

martijnjanssen:
Looks good!

PR #313, opened 2016-11-25
Remove calc statements for footer position

martijnjanssen:
@TimvdLippe I think you should take a look at this :smile:

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,6 +51,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   <style>
     body {
+      min-height: 100vh;
+      position: relative;
       background: #fafafa;
       font-family: 'Roboto', 'Helvetica Neue', Helvetica, Arial, sans-serif;
       color: #333;
--- a/index.html
+++ b/index.html
@@ -51,8 +51,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   <style>
     body {
-      min-height: 100vh;
-      position: relative;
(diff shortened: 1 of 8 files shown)


martijnjanssen:
@TimvdLippe

TimvdLippe:
Hm it does not seem to solve the issues we have. If we make the footer absolute, it either makes the page too long (for small pages) or is on top of content of long pages. :sob:

TimvdLippe:
Shall we just close this one? If it ain't broke, don't fix it 😜

martijnjanssen:
I'm going to adjust it to only have a minimal height above a certain width.

martijnjanssen:
Unfortunately, the footer position will not work for some strange reason with the correct styling properties. [This](https://codepen.io/anon/pen/PbvpKK?utm_campaign=chrome_series_codesample_111516&utm_source=chromedev&utm_medium=yt-desc) example makes it look so easy, but for some reason I couldn't get it to work for our page. The footer would just float at the `position: relative; bottom: 0;` position.

martijnjanssen:
Okay, It is finally fixed, the footer is now properly positioned at the bottom at all times! The only sad thing is that I can't force the `contentContainer` in the `app-header-layout` to have `min-height: 100%` due to the fact that there is no css variable and the selector does not work. This is the cause of the only needed `calc(100vh - 64px)` statement. If that issue is resolved, the page will have no calc statements for the height, although the header is not going to change in size. Any thoughts on this?

PR #314, opened 2016-11-25
Simplify lancie-card & remove unused styling

martijnjanssen:


src/lancie-card.html
```diff
@@ -4,29 +4,25 @@
...
+        }
+        --paper-card-header-image: {
+          min-height: 120px;
+          background-color: #1a2b43;
         }
       }
     </style>
 
-    <paper-card heading="{{heading}}" image="https://areafiftylan.nl/images/logos/lancie/logo_without_date_with_background.png">
+    <paper-card heading="[[heading]]" image="images/logos/lancie/logo_without_date_with_background.png">
```
TimvdLippe: Shouldn't this be `../images` since the images is in the parent folder of `src/`

PR #316, opened 2016-11-27
Extract login form & add flag to bypass login

martijnjanssen:
~~I could use your help here @TimvdLippe~~

~~I have extracted the login form and made sure all the buttons work. The only thing I can't figure out is how to correctly add a flag to a `lancie-login-check` method to bypass the login check and make a page public. I have already added a variable and have set `nologin` variable in the `lancie-content` on the ticket-page login-check.~~

The logic is working as intended now, adding a `nologin` variable to the login-check in lancie-content now allows the user to view the page without logging in. We may need to rename the lancie-login-check because this functionality now bypasses the checking behavior. 

Would `lancie-login-catch` be a better name?

### DIFF
--- a/src/lancie-login/lancie-login-check.html
+++ b/src/lancie-login/lancie-login-check.html
@@ -1,13 +1,13 @@
-<link rel="import" href="lancie-login-form.html">
+<link rel="import" href="lancie-login-card.html">
 
 <dom-module id="lancie-login-check">
   <template>
-    <template is="dom-if" if="{{exists(user)}}" restamp="true">
+    <template is="dom-if" if="{{canAccess(user, nologin)}}" restamp=true>
       <content></content>
     </template>
 
-    <template is="dom-if" if="{{!exists(user)}}">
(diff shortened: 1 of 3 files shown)


src/lancie-login/lancie-login-check.html
```diff
@@ -29,16 +29,26 @@
           value: function() {
             return user;
           }
+        },
+        nologin: {
+          type: Boolean,
+          value: function() {
```
TimvdLippe: Can just be `false`

src/lancie-login/lancie-login-check.html
```diff
@@ -29,16 +29,26 @@
...
         }.bind(this));
       },
 
-      exists: function(value) {
-        return value !== '';
+      canAccess: function(user, nologin) {
+        if (nologin) {
+          return true;
+        }
+        return user !== '';
```
TimvdLippe: Can be `nologin || !user`

PR #320, opened 2016-11-30
Update serve command to use the renamed polyserve arguments

TimvdLippe:
`npm run serve` once again works when initially cloning the repository.

Make sure to run `npm update && npm install` to fetch the updated version of `polyserve`.


martijnjanssen:
Looks good! Thanks!

PR #323, opened 2016-12-02
Feature ticket page rework

### LINKED ISSUE #230: Open order is not loaded

### LINKED ISSUE #257: Can't accept terms & conditions on portrait screens.

### LINKED ISSUE #272: Account details intercept flow
Currently, we check the profile details before the order creation. This needs to be done on checkout. Preferably in a dialog before the T&C.

### LINKED ISSUE #344: Refactor lancie-ticket-block
During the AV, someone suggested for the Early bird tickets to not have a set number, but a time limit instead. After our meeting, we decided that we want that too. I checked throughout our system, and the implementation is already there, except in this element early birds cannot have a deadline. Refactoring this will also help removing the hardcoded tickettypes, and moving to dynamic tickets.

The new code should check if either the deadline, the limit, or both are set, and handle accordingly.

martijnjanssen:
Requires https://github.com/AreaFiftyLAN/lancie-api/pull/314

Fixes #344 
Fixes #272 
Fixes #257 
Fixes #230 

~~So right now it is not working as intended. In `lancie-login-check` the `this.$` selector seems to be broken for no reason. Could you look at this @TimvdLippe ?~~


src/lancie-login/lancie-login-check.html
```diff
@@ -40,6 +44,18 @@
         checks.push(function(newUser) {
           this.user = newUser;
         }.bind(this));
+
+        // Add eventlistener to catch logged-in event in entire element
+        this.addEventListener('logged-in', (e) => {
```
TimvdLippe: Change to `function(e) {` ES5 syntax

src/lancie-ticket-page/lancie-ticket-basket.html
```diff
@@ -0,0 +1,89 @@
...
+
+<link rel="import" href="../lancie-login/lancie-ajax.html">
+<link rel="import" href="lancie-ticket-basket-item.html">
+
+<dom-module id="lancie-ticket-basket">
+  <template>
+	<style>
+	</style>
+
+    <lancie-ajax id="removeTicket" refurl="/orders/[[basket.id]]" method="DELETE" on-response="onRemoveTicket"></lancie-ajax>
```
TimvdLippe: It seems that the indentation is off in this file.

TimvdLippe: There seems to be tab characters in this file, instead of 2 spaces

MatthijsKok:
#344 can nicely be done as well.

martijnjanssen:
@TimvdLippe could you look at this?

skamoen:
Clicktesting issues 1

1. Create an order without logging in
2. Login with existing user
3. Go through mollie
4. Callback to `/ordersuccess?order=...` results in an error page

The X-Auth-Token header is empty, therefore the orderstatus call results in 403

skamoen:
Clicktesting issue 2

1. Create an order without loggin in
2. Register a new user
3. Click the link in the mail

If there's an orderId in the url, you can assume the user wants to finish the order.

skamoen:
Clicktesting issue 3

Profile intercept form:
- 409 Profile Name in use error isn't handled, shows as unable to connect in the background. You want these errors more visible
- Succesful submit doesn't close the dialog. Dialog has to be closed and needs another click on checkout before I can continue.

src/lancie-login/lancie-login-form.html
```diff
@@ -128,8 +128,11 @@
         }
       },
 
-      tryRegister: function() {
+      tryRegister: function(id) {
         if (this.validateRegisterForm()) {
+          if (!!id) {
+            this.$.ajaxRegister.body['orderId'] = id;
```
TimvdLippe: This feels very wrong. Maybe expose the `body` property so that users of the login-form can specify which body to send?

src/lancie-login/lancie-login-form.html
```diff
@@ -170,12 +173,11 @@
...
         });
-     },
+      },
 
-      onRegister: function(e, response) {
-        if (response.succeeded) {
-          this.fire('navigate', {page: 'check-email'});
-        }
+      onRegister: function(e, request) {
+        this.fire('navigate', {page: 'check-email'});
```
TimvdLippe: The condition for `response.succeeded` has been removed. Intended?

src/lancie-login/lancie-login-form.html
```diff
@@ -191,7 +193,7 @@
       bodyFrom: function(username, password) {
         return {
           username: username,
-          password: password
+          password: password,
```
TimvdLippe: Extra comma is unnecessary, but does not harm.

src/lancie-login/lancie-register-confirmation.html
```diff
@@ -44,26 +44,25 @@
...
       },
 
       attached: function() {
         this.confirming = true;
         this.success = false;
         this.error = false;
         this.tokenTwice = false;
-        this.token = window.location.href.split('token=')[1];
-        this.tryRequestPasswordReset();
+        this.search = window.location.search;
```
TimvdLippe: This can change over time. You can also use `<iron-location>` with the `query` property: https://beta.webcomponents.org/element/PolymerElements/iron-location/iron-location#property-query

src/lancie-login/lancie-register-confirmation.html
```diff
@@ -75,7 +74,7 @@
             this.tokenTwice = true;
           }
         }
-      }
+      },
```
TimvdLippe: Not explicitly required.

src/lancie-my-area/profile-edit-form.html
```diff
@@ -1,4 +1,9 @@
 <link rel="import" href="../../bower_components/iron-flex-layout/iron-flex-layout-classes.html">
+<link rel="import" href="../../bower_components/paper-item/paper-icon-item.html">
+<link rel="import" href="../../bower_components/paper-dropdown-menu/paper-dropdown-menu.html">
+<link rel="import" href="../../bower_components/iron-icons/communication-icons.html">
+<link rel="import" href="../../bower_components/iron-icons/social-icons.html">
+<link rel="import" href="../../bower_components/paper-input/paper-input.html">
```
TimvdLippe: :+1: :+1: :+1: :+1: :+1: :+1:

src/lancie-my-area/profile-edit-form.html
```diff
@@ -53,7 +58,6 @@
         <paper-input label="Phone Number" type="text" value="{{profile.phoneNumber}}"></paper-input>
       </div>
     </div>
-    <paper-icon-item>
```
TimvdLippe: :joy: Oops

src/lancie-ticket-page/lancie-profile-dialog.html
```diff
@@ -49,24 +45,24 @@
 
     <lancie-ajax id="ajaxUpdateProfile" refurl="users/current/profile" method="POST" on-response="onProfileUpdate"></lancie-ajax>
 
-    <lancie-card heading="Fill in your profile">
+    <paper-dialog id="dialog" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
```
TimvdLippe: Shouldn't this be `lancie-dialog`?

src/lancie-ticket-page/lancie-profile-dialog.html
```diff
@@ -80,14 +76,14 @@
           type: Boolean,
           value: true
         },
-        route: {
-          type: String,
-          notify: true
-        }
+      },
+
+      ready: function() {
```
TimvdLippe: `ready` => `attached`?

TimvdLippe: Also on `detached` we probably need to remove the dialog again.

src/lancie-ticket-page/lancie-ticket-basket-item.html
```diff
@@ -61,10 +60,10 @@
       removeTicket: function() {
         this.fire('remove-ticket', {
           type: this.ticket.type.name,
-          options: this.ticket.enabledOptions.map(function(option) {
+          options: this.ticket.enabledOptions.map(function       (option) {
```
TimvdLippe: Unnecessary whitespace

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -31,102 +32,120 @@
...
+            value: [],
+          },
         },
-        basket: {
-          type: Object,
-          value: {},
-          // notify: true,
-          observer: '_basketChanged',
+
+        ready: function () {
```
TimvdLippe: `attached`

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -31,102 +32,120 @@
...
+          }
+        },
+
+        onCoupleOrder: function (e, request) {
+          if (request.succeeded) {
+            this.basket = request.response.object;
+            this.$.orderStorage.removeStorage();
+          } else {
+            this.basket.status = 'ANONYMOUS';
+            this.fire('toast', {text: 'Unable to couple the order to your account. Try refreshing the page.'});
```
TimvdLippe: Instead of refreshing, can we maybe fix the status?

TimvdLippe:
I have updated the PR description to correctly reference multiple issues to be fixed

skamoen:
clicktesting issue:

When coming from the confirm email (with orderId in url), the confirmRegistrationpage should say "Click here to continue your order". That link goes to the `/tickets` page, but should also automatically trigger the popup window, for example with `/tickets?login`

TimvdLippe:
Opening https://127.0.0.1:5100/tickets and logging throws a 401 error. Subsequently adding tickets to the order shows the toast "Unabled to connect to the server". Refreshing the page fixes this issue.

TimvdLippe:
Adding a ticket with no ticket options selected and a ticket with all options selected, deleting the ticket with no options selected sometimes removes the other ticket.

src/lancie-content.html
```diff
@@ -166,7 +166,7 @@
       },
 
       _filterConstruction: function(page) {
-        if (window.location.host.indexOf('localhost') !== -1) {
+        if (window.location.host.indexOf('localhost') >= 0 || window.location.host.indexOf('127.0.0.1') >= 0) {
```
martijnjanssen: Shouldn't this be `> -1`?

TimvdLippe: Logic is equivalent. You can do so, doesn't really matter I think

src/lancie-login/lancie-login-form.html
martijnjanssen: This is done because this function is called on the `response` event. If the request was not succeeded, the `error` event would be fired, correct?

src/lancie-ticket-page/lancie-profile-dialog.html
martijnjanssen: It could be, but I am still hesitant to do so, I am still not quite sure what is going to happen to the current `lancie-dialog` element. Are we going to update the element or are we going to build a new `lancie-dialog`?

src/lancie-login/lancie-login-form.html
martijnjanssen: So outside of the element you retrieve some data from the element and then set the body with the data you retrieved from the element?

martijnjanssen:
Did you mean 'logging in' in [this](https://github.com/AreaFiftyLAN/lancie-frontend/pull/323#issuecomment-269988403) comment?

src/lancie-login/lancie-login-form.html
TimvdLippe: The user of lancie-login-form creates the body, not the form itself.

src/lancie-login/lancie-login-form.html
TimvdLippe: Not sure. Probably is but not 100% sure

martijnjanssen:
I think this is the final version of the new ticket page. @TimvdLippe what do you think of the fixes?

I have some doubts about the extraction of the methods in login-form. I would like to change the login form to a iron-form soon. But this will not be possible when the methods are outside of the form itself. For instance submit on enter will not work in that case.

MatthijsKok:
Email check happens after page is unloaded.
- Make sure you're not logged in.
- Click on My Area.
- Click on Seatmap.
- Click on My Area again.
- Observe that the validation happened, without ever typing something.

martijnjanssen:
You are correct, this is a thing that happens, but this is due to the fact that chrome has autofocus on the username field. If you observe what happens you see that the field is focussed on seeing the username field. When changing the page to another one, focus is lost. This is called the 'blur' event, which triggers validation of the field. If you want to change anything about this, please open an issue, or even better, a PR. But again, this is not an issue introduced by this PR. This won't be fixed here.

src/lancie-ticket-page/lancie-ticket-item.html
```diff
@@ -0,0 +1,232 @@
...
+
+        this.options = options;
+        this.available = this._available();
+      },
+
+      addToBasket: function() {
+        if (!this.basket.tickets) {
+          this.$.addTicket.refurl = 'orders';
+        } else if (this.basket.tickets.length >= 5) {
+          this.fire('toast', {text: 'Maximum of 5 tickets.'});
```
MatthijsKok: Would like this changed to: "A maximum of 5 tickets can be added."

MatthijsKok: "Maximum of 5 tickets in basket." is even better

src/lancie-ticket-page/lancie-ticket-basket.html
```diff
@@ -0,0 +1,160 @@
...
+
+  (function() {
+	Polymer({
+	  is: 'lancie-ticket-basket',
+	  properties: {
+      user: Object,
+		  basket: {
+			  type: Object,
+			  notify: true,
+		  },
```
TimvdLippe: Indentation is off here

src/lancie-ticket-page/lancie-ticket-basket.html
```diff
@@ -0,0 +1,160 @@
...
+      if (!this.$.profileDialog.allFilledIn(this.user)) {
+        this.$.profileDialog.open();
+      } else {
+        this.$.termsOfService.open();
+      }
+    },
+
+    onCheckout: function(e, request) {
+      if (request.succeeded) {
+        window.location.href = request.response.object;
```
TimvdLippe: Add a comment this redirects to Mollie, so it is not an internal url change

src/lancie-ticket-page/lancie-ticket-item.html
```diff
@@ -0,0 +1,232 @@
...
+        },
+        color: String,
+      },
+      observers: [
+        '_calulatePrice(options.*)'
+      ],
+
+      ready: function() {
+        var options = [];
+        for (var option in this.ticket.possibleOptions) {
```
TimvdLippe: Ticket might be undefined here? Better to add it in an observer

src/lancie-ticket-page/lancie-ticket-item.html
```diff
@@ -0,0 +1,232 @@
...
+          return 'Available until: ' + new Date(this.ticket.deadline).toLocaleString('nl', {hour12: false});
+        } else if (deadlineType === 'limit') {
+          return 'Only ' + (ticket.limit - ticket.numberSold) + ' left!';
+        } else {
+          return 'Available';
+        }
+      },
+
+      _calulatePrice: function(changed) {
+        if (!this.ticket) {
```
TimvdLippe: Add this as argument to the observer, makes sure ticket is never undefined

src/lancie-ticket-page/lancie-ticket-item.html
```diff
@@ -0,0 +1,232 @@
...
+        classes += (available) ? '' : ' sold-out';
+        return classes;
+      },
+
+      _hasOptions: function(options) {
+        return options.length > 0;
+      },
+
+      _available: function() {
+        var deadlineType = this.deadlineDisplay[this.ticket.ticketType];
```
TimvdLippe: Also has a dependency on ticket

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -4,263 +4,154 @@
...
+              this.fire('toast', {text: 'Unable to couple the order to your account. Try refreshing the page.'});
+            }
+          }
+        },
+
+        /*
+          If there is an order and the user logs in later, couple the order to the user.
+         */
+        _userChanged: function (user) {
+          if (this.basket.status === 'ANONYMOUS' && !!user) {
```
TimvdLippe: Dependency on basket

src/lancie-ticket-page/lancie-tickets.html
```diff
@@ -0,0 +1,70 @@
...
+              'green',
+              'purple',
+              'orange',
+            ];
+          },
+        },
+      },
+
+      _getColorClass: function(index) {
+        return this.colors[index];
```
TimvdLippe: What if index is greater than the length?

PR #326, opened 2016-12-14
Remove moment.js and unused activity-footer element

TimvdLippe:
Moment.js was only used in this place and this element was actually never used anyways: `data.footer` is always false. Therefore purge the element and the dependency :tada:


martijnjanssen:
Looks great, always nice to remove unneeded code and dependencies!

PR #327, opened 2016-12-14
Lazy load polyfill only on browsers without native custom elements

TimvdLippe:
The polyfill is now only loaded in browsers without native custom elements, such as Firefox. It is not loaded in Chrome.


martijnjanssen:
Nice, this fixes the unneeded request!

PR #329, opened 2016-12-14
Optimize images with ImageOptim

TimvdLippe:
You can run images optimized with `npm run build optimize-images`

Current live website size: 8.4 MB
Generated build locally: 2.8 MB


martijnjanssen:
Thanks for the fix! The webpage is way faster now!

PR #330, opened 2016-12-15
Navigate pages by chaging the page in lancie-content

martijnjanssen:


src/lancie-content.html
```diff
@@ -128,6 +128,9 @@
         this.addEventListener('store-token', function(e) {
           this.$.auth.storeToken(e.detail.token);
         }.bind(this));
+        this.addEventListener('navigate', function(e) {
+          this.set('page.page', e.detail.page);
```
TimvdLippe: General comment: could just have been `e.detail`. E.g. instead of passing an object to `this.fire` just directly pass the String.

PR #333, opened 2016-12-19
Enable shadow dom and fix incompatibility with previous implementation

### LINKED ISSUE #128: Consider enabling shadow-dom

TimvdLippe:
This version should now be compatible with both shady and shadow dom.
Most prominently lancie-activity-dialog had to be rewritten as it
was relying on illegal ::shadow selectors and styling

Fixes #128


martijnjanssen:
It seems like the dom-if contents in the login-catch are rendered since the requests are showing up in my console. This should not happen as far as I know.

martijnjanssen:
It also looks like something in the my-area styling changed. The title of the tickets card looks like it is missing some styling. 
image

martijnjanssen:
While in the seatmap, just opening and closing seats for some time seems to slow down my browser incrementally. In the task manager you see the usage going to 100% for some reason.

TimvdLippe:
The timing issue has been fixed, but it required using `iron-lazy-pages#2.0-preview`. The styling issue still has to be fixed, but not sure what we want to fix there.

TimvdLippe:
Will wait till #323 is merged and then I will rebase this PR

TimvdLippe:
> Why do some elements in lancie-content have a template and some do not?

Those inside a `login-check` are in a template, such that the template is stamped when the route is shown. This prevents the incorrect invocation of `attached`

martijnjanssen:
The styling issue is now fixed, but the data binding is broken in my-area and some other pages.

TimvdLippe:
> the data binding is broken in my-area and some other pages

Could you please be more specific? Have some time to debug now :tada:

martijnjanssen:
Okay, I have checked yesterday and added some observers on for instance the `{{user}}` binding from content into `my-area`. The observer in `lancie-content` does get the user, but this value is not reflected in the user variable in `my-area`.

MatthijsKok:
Since #323 is now merged, this is no longer blocked.

TimvdLippe:
Rebased on top of current master. This PR can be once again reviewed and merged :smile:

martijnjanssen:
Nice, great stuff!

PR #335, opened 2016-12-21
Construction page

martijnjanssen:
@TimvdLippe now only figure out how to change the route to this on specific pages.

### DIFF
--- a/src/lancie-content.html
+++ b/src/lancie-content.html
@@ -123,6 +123,7 @@
         route: {
           type: String,
           notify: true,
+          observer: '_changed',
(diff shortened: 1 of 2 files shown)


src/lancie-content.html
```diff
@@ -106,6 +103,14 @@
         <lancie-contact-success></lancie-contact-success>
       </template>
 
+     <template is="iron-lazy-page" data-route="construction" path="src/lancie-coming-soon.html">
```
TimvdLippe: Indentation ;)

PR #338, opened 2016-12-21
Remove import breaking contact form

martijnjanssen:
This import is not needed and results in actually breaking when changing to the actual contact form page. It tries to register the contact form page but is unable to because the dom-module already exists. Looks like this only happens in production. Is this interesting to you @TimvdLippe?


skamoen:
@martijnjanssen fault if it breaks.

PR #345, opened 2017-01-02
Fire requests only with token & encapsulate js

martijnjanssen:
Some requests are failing because they are sent in the attached function. This results in some headers not being set and the requests failing for valid reason. This changes the firing method to wait for the headers to be correctly set.


TimvdLippe:
What is the reason for these changes?

martijnjanssen:
@TimvdLippe I assume you wanted a description for the PR, I have updated the description to explain the reason for these changes.

PR #352, opened 2017-01-04
Separate error element

martijnjanssen:
Completely removed all the duplicate code and recurring code. This has removed quite some code from all elements and allowing for an easier way of showing errors.

### DIFF
--- a/src/lancie-error/lancie-error.html
+++ b/src/lancie-error/lancie-error.html
@@ -21,9 +21,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         background: #FFCDD2;
         color: #B71C1C;
       }
+
+      [hidden] {
+        display: none !important;
+      }
     </style>
 
-    [[message]]
+    <span>[[message]]</span>
 
   </template>
   <script>
--- a/src/lancie-error/lancie-error.html
+++ b/src/lancie-error/lancie-error.html
@@ -1,0 +1,57 @@
+<!--
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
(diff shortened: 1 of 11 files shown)


src/lancie-error/lancie-error.html
```diff
@@ -0,0 +1,57 @@
...
+        display: block;
+        width: inherit;
+        padding: 10px 20px;
+        border-left: 4px solid #F44336;
+        background: #FFCDD2;
+        color: #B71C1C;
+      }
+    </style>
+
+    [[message]]
```
TimvdLippe: Wrap this in a `<span>`

src/lancie-error/lancie-error.html
```diff
@@ -0,0 +1,57 @@
...
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<link rel="import" href="../../bower_components/polymer/polymer.html">
+
+<dom-module id="lancie-error">
+  <template>
+    <style>
+      :host {
```
TimvdLippe: Needs extra rule to actually `display: none` on hidden. This is not propagated in shadow dom

src/lancie-login/lancie-login-form.html
```diff
@@ -109,50 +96,37 @@
...
         this._token = request.response.object;
         this.$.getCurrentUser.generateRequest();
       },
 
-      onFailedLogin: function(event, request) {
-        this.loginError = true;
+      onLoginError: function(event, request) {
         if (request.request.status === 401) {
-          this.loginErrorMessage = 'Bad credentials.';
+          this.$.error.setError('Bad credentials.')
```
skamoen: While we're doing stuff here, please add an extra catch for unverified users. The request message will be "Authentication Failed: User is disabled", which should display an error like "Please validate your email first."

martijnjanssen: I will fix this in the PR fixing the login endpoint.

PR #356, opened 2017-01-05
Add password reset & improve logic

martijnjanssen:


MatthijsKok:
- The register screen has a forgot password, which i don't like.
- I would like to see a "Back to login" button on the password reset screen.
- "Forgot password" itself doesn't look perfect as is... Maybe align it to the right?

martijnjanssen:
Whoops, thanks for noticing that first one! Completely missed that one.

PR #360, opened 2017-01-06
Update teams size in tournaments and add logo in activity window on homepage

Hvanderwilk:
Fixed some incorrect information and added pictures where none were present.

### DIFF
--- a/scripts/activity.json
+++ b/scripts/activity.json
@@ -7,7 +7,7 @@
     "backgroundImagePath": "images-optimized/activities/rl.jpg",
     "sponsor": {
       "name": "",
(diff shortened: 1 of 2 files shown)


martijnjanssen:
The images are the places where the sponsor logos will eventually be added. I don't think the dialog needs the same image from the game.

Hvanderwilk:
Half of the tournaments had images of the game logos, that's why I changed the others. Maybe a "Your logo here?" picture will suit better.

martijnjanssen:
Let's not do that as discussed.

Could you also give the json location in the homepage a version bump? A minor version increase would be fine.

PR #361, opened 2017-01-06
Fix ticket transfer & simplify element

### LINKED ISSUE #357: TicketTransfer bugs.
- [x] Popup doesn't dissappear when ticket is transferred.
- [x] Popup doesn't dissappear when ticket transfer is cancelled.
- [x] No succes notification happens for both actions.
- [x] When a ticket is already setup for transfer, the API sends `Ticket XX is already set up for transfer!` but the frontend just displays `The server was unable to process your request.`

martijnjanssen:
Fixes #357 
Reworked and simplified the transfer and cancel flows to improve on readability and functionality.

### DIFF
--- a/src/lancie-my-area/my-area-orders.html
+++ b/src/lancie-my-area/my-area-orders.html
@@ -60,7 +60,7 @@
       <div class="card-content">
         <template is="dom-if" if="[[tickets.length]]">
           <template is="dom-repeat" items="[[tickets]]" index-as="index">
-            <my-area-ticket-item username="{{username}}" ticket="[[item]]"></my-area-ticket-item>
+            <my-area-ticket-item user="[[username]]" ticket="[[item]]"></my-area-ticket-item>
           </template>
         </template>
         <template is="dom-if" if="[[!tickets.length]]">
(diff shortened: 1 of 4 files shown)


MatthijsKok:
- [ ] "Email should be valid." check is done immediately.
- [ ] You can send a ticket to yourself. (API bug probably)

Otherwise great changes!

MatthijsKok:
`"Email should be valid." check is done immediately.`
I think this needs to get fixed after all, Since the `label` already states `Email`, so it is not unclear to the user what to enter in the field.

MatthijsKok:
I did not get any error when trying to transfer the ticket to myself. Maybe this is broken because the API now throws an exception?

martijnjanssen:
The api fix was just pushed, this is not included in this PR.

src/lancie-my-area/my-area-orders.html
```diff
@@ -57,7 +57,7 @@
       <div class="card-content">
         <template is="dom-if" if="[[tickets.length]]">
           <template is="dom-repeat" items="[[tickets]]" index-as="index">
-            <my-area-ticket-item username="{{username}}" ticket="[[item]]"></my-area-ticket-item>
+            <my-area-ticket-item user="[[username]]" ticket="[[item]]"></my-area-ticket-item>
```
TimvdLippe: Shouldnt this be user?

PR #362, opened 2017-01-06
Rename file and routes to actions

martijnjanssen:
The routes are now renamed to the actions they take or represent.


TimvdLippe:
Let's rename after merging #333 . Other than that LGTM

PR #363, opened 2017-01-07
Lancie-form to ease validation & sending request

### LINKED ISSUE #54: Profile updating needs better error handling
I would like to see the editing of the profile data a little different. When pressing save and all went right is perfect right not. When errors occur I would like to still see the error message. Besides that the form should not disappear but should stay, all fields that have an error should color red and show a small message (popup or below/above) showing what went wrong.

martijnjanssen:
Fixes #54 

This should it make it even more easier to require user input. It takes input and turns it into a request. Some things still need to be done, like:
- [x] Add the token to requests

### DIFF
--- a/src/lancie-my-area/my-area-profile.html
+++ b/src/lancie-my-area/my-area-profile.html
@@ -22,6 +22,10 @@
       display: block;
     }
 
+    [hidden] {
+      display: none !important;
+    }
+
     paper-tooltip {
       --paper-tooltip-background: var(--primary-color);
       --paper-tooltip-text-color: var(--secondary-color);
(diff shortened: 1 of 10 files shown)


src/lancie-my-area/my-area-profile.html
```diff
@@ -102,60 +102,44 @@
...
-            <iron-icon icon="icons:lock" item-icon></iron-icon>
-            <div class="flex">
-              <paper-input label="Current Password" type="password" value="{{currentPassword}}"></paper-input>
-              <paper-input label="New Password" type="password" value="{{newPassword}}"></paper-input>
-              <paper-input id="passwordRepeatInput" error-message="Passwords don't match!" label="Repeat New Password" type="password" value="{{newPasswordRepeat}}"></paper-input>
-            </div>
+        <div class="layout horizontal center" hidden$="[[editingPassword]]">
+          <iron-icon icon="icons:lock" item-icon></iron-icon>
+          <div class="item flex">
+            <div>&#9679;&#9679;&#9679;&#9679;&#9679;&#9679;&#9679;&#9679;&#9679;&#9679;&#9679;&#9679;</div>
```
TimvdLippe: Um?

martijnjanssen: Those are the password bullets shown when editing a password. I have not changed this.

FrancisBehnen:
#### Safari
- tab and enter navigation works :)
- Surprise surprise, alignment is horrible 
screen shot 2017-01-08 at 20 24 55
- You can also see that the icons are tiny, this problem exists in Chrome as well (not in Firefox). This is because the icons scale with the width of the website..

FrancisBehnen:
Safari alignment is not a problem of this PR, flex wrap doesn't work in anywhere.. https://github.com/PolymerElements/iron-flex-layout/issues/35

FrancisBehnen:
Looks alright to me now

MatthijsKok:
image
this is how my popup looks. (buttons are way smaller)
image
and this is my normal login form.

MatthijsKok:
LGTM otherwise.

PR #364, opened 2017-01-07
New lancie-dialog

FrancisBehnen:
Still needs:
- [x]  Dismiss functionality
- [x]  Dimmed background 
- [x]  Remove `lancie-dialog-old.html`


FrancisBehnen:
@martijnjanssen, any idea how I should get the `dialog-dismiss` button to actually close the dialog?

FrancisBehnen:
@TimvdLippe, @martijnjanssen, any thoughts on this `paper-dialog` in our custom dialog. Way to go, or do we want something else?

FrancisBehnen:
Middle of dialog is not in the middle of the page, but the positioning looks good to me anyway.

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,74 +1,120 @@
...
+     is: 'lancie-dialog',
+
+     listeners: {
+       'tap': 'isDismissButton'
+     },
+
+    /**
+     * Find out if the tapped place was within the dismiss button
+     */
+     isDismissButton: function(event) {
```
TimvdLippe: Instead of trying to find out if the button was tapped, I think it is better to fire an event `dismiss` inside the element passed onto `<content>`, listen there to `tap` and fire the `dismiss` event

MatthijsKok:
Tested and working in Chrome 55 and Firefox 46.

FrancisBehnen:
@martijnjanssen and I decided to switch to a `paper-dialog` implementation since the cpu increase he observed was not caused by the usage of `paper-dialog` features and this makes the implementation easier and cleaner

MatthijsKok:
image
this is how my popup looks. (buttons are way smaller)
image
and this is my normal login form.

FrancisBehnen:
In response to Matthijs' comment, do we want full width buttons? 
@martijnjanssen I'll remove it as soon as it's finished, I still use it to compare to the new dialog.

martijnjanssen:
The full-width buttons are fine with me, but ONLY in the login dialogs. As for the other dialogs, the buttons shouldn't even be raised. The actions should be flush with the card-actions.

image
image

With these two images, the bottom dialog has the raised buttons. The buttons alone look nice, but it is a cleaner and simpler design with the buttons flush on the card. I would prefer a simple and cohesive design over a design with all kinds of fancy unneeded additions.

FrancisBehnen:
Decided with Tim that we don't want `paper-dialog` in the end.. It's too much of a hassle and it breaks all kinds of weird things..

FrancisBehnen:
Do we want to invest time in making the animations work again? I don't think it's worth the effort, so I vote no.
@MatthijsKok, @martijnjanssen, @TimvdLippe, @skamoen, @Hvanderwilk

TimvdLippe:
I dont think the animations add value, so we can do without

src/lancie-dialog/lancie-dialog.html
FrancisBehnen: @MatthijsKok suggested the same today. I don't like this function either, but I would like `lancie-dialog` to be able to handle such a simple thing as dismissing itself without requiring additional logic in the element that uses `lancie-dialog`.

FrancisBehnen: Or did you mean that `lancie-dialog` should add the `tap` listener to the button?

FrancisBehnen:
Any more requests? Otherwise this is `ready for review`

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,74 +1,134 @@
...
-          this._finishRenderClosed();
+
+     is: 'lancie-dialog',
+
+     listeners: {
+       'tap': 'isDismissButton'
+     },
+
+    /**
+     * Find out if the tapped target was the dismiss button
```
TimvdLippe: This is a hack, there must be a better solution. I will fix this tomorrow so you can see the difference :) Very hard to explain otherwise.

FrancisBehnen:
Consider merging #395 into this PR first

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,72 +1,177 @@
...
-        if (this.withBackdrop) {
-          this.backdropElement.open();
+
+      addDismissListeners: function() {
+        var host = this;
+        var dismissEls = this.querySelectorAll('[dialog-dismiss]');
+        for (var i = 0; i < dismissEls.length; i++) {
+          dismissEls[i].addEventListener('tap',
+            function() { host.close(); });
         }
```
martijnjanssen: I would change these lines to:
``` javascript
dismissEls[i].addEventListener('tap', function() {
    host.close();
});
￼```

File: src/lancie-dialog/lancie-dialog.html [hunk 4]
```diff
@@ -1,72 +1,177 @@
...
+        this.addDismissListeners();
       },
-      _renderOpened: function() {
-        this.cancelAnimation();
-        if (this.withBackdrop) {
-          this.backdropElement.open();
+
+      addDismissListeners: function() {
+        var host = this;
+        var dismissEls = this.querySelectorAll('[dialog-dismiss]');
```
martijnjanssen: I would change the variable name to `dismissButtons`

FrancisBehnen:
Blocked until everyone has taken his time to review this.

martijnjanssen:
It looks like there is an issue with the margin in lancie-activity-dialog combined with the scrolling. I believe this has to do with the flex display that is put on the `.card-content` in the `lancie-activity-dialog.html`. Could you please take a look at the way things are styled in there? Looks like it could be improved since it also has a duplicate media query.
image

I will now take a look at the code-side of things :+1:

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,72 +1,177 @@
...
+       * with `--lancie-dialog-max-height: calc(--app-toolbar-height * 10);`
+       * but no variable support in @media queries (yet)
+       */
+      @media (max-height: 640px) {
+        #actualDialog {
+          @apply(--lancie-dialog-max-height-640);
+        }
+      }
+
+      @media (max-height: 213px) {
```
martijnjanssen: This media query looks very redundant to me.

src/lancie-login/lancie-login-dialog.html
```diff
@@ -12,61 +12,31 @@
...
-    <paper-dialog id="dialog" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
+    <lancie-dialog id="dialog">
       <h2>[[_getAction(registering)]]</h2>
-      <lancie-login-form id="loginForm" registering="[[registering]]" order-id="[[orderId]]" on-enter="submit"></lancie-login-form>
+      <lancie-login-form class="card-content" id="loginForm" registering="[[registering]]" order-id="[[orderId]]" on-enter="submit"></lancie-login-form>
 
-      <div class="buttons dialog-actions">
-        <paper-button class="btn btn-primary" raised on-tap="submit">[[_getAction(registering)]]</paper-button>
-        <paper-button class="btn" raised on-tap="switchForm">[[_getSwitchAction(registering)]]</paper-button>
+      <div class="buttons full-width">
```
martijnjanssen: The `full-width` class is never used.

src/lancie-my-area/my-area-orders.html
```diff
@@ -75,15 +73,9 @@
...
-      <h2>Terms and conditions</h2>
-      <paper-dialog-scrollable>
-        <lancie-terms-of-service-text></lancie-terms-of-service-text>
-      </paper-dialog-scrollable>
-      <div class="buttons">
-        <paper-button dialog-dismiss>Close</paper-button>
-      </div>
-    </paper-dialog>
+    <lancie-terms-of-service id="termsOfService">
+      <paper-button on-tap="closeDialog">Close</paper-button>
```
martijnjanssen: Can't this button have a `dialog-dismiss` attribute?

src/lancie-ticket-page/lancie-profile-dialog.html
```diff
@@ -37,20 +36,21 @@
...
         background: var(--primary-color);
         margin: auto;
+        margin-bottom: 10px;
       }
     </style>
 
     <lancie-ajax id="ajaxUpdateProfile" refurl="users/current/profile" method="POST" on-response="onProfileUpdate"></lancie-ajax>
 
-    <paper-dialog id="dialog" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
+    <lancie-dialog id="dialog" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
```
martijnjanssen: Animations can be removed.

martijnjanssen: `with-backdrop` can also be removed, not using it.

src/lancie-ticket-page/lancie-terms-of-service.html
```diff
@@ -1,27 +1,44 @@
...
+
+        --lancie-dialog-max-height-640: {
+          @apply(--lancie-terms-of-service-fullscreen);
+        };
+      }
+      
     </style>
 
-    <paper-dialog id="termsOfService" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
+    <lancie-dialog id="termsOfService" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
```
martijnjanssen: Unneeded attributes can be removed.

src/lancie-ticket-page/lancie-ticket-basket.html
```diff
@@ -83,7 +83,10 @@
 
     <lancie-profile-dialog id="profileDialog" user="{{user}}"></lancie-profile-dialog>
 
-    <lancie-terms-of-service id="termsOfService" checkout="[[checkout]]"></lancie-terms-of-service>
+    <lancie-terms-of-service id="termsOfService">
+      <paper-button on-tap="closeDialog">Cancel</paper-button>
+      <paper-button on-tap="checkout">I agree</paper-button>
```
martijnjanssen: I would change 'I agree' to 'Agree'.

src/lancie-ticket-page/lancie-ticket-basket.html
```diff
@@ -83,7 +83,10 @@
 
     <lancie-profile-dialog id="profileDialog" user="{{user}}"></lancie-profile-dialog>
 
-    <lancie-terms-of-service id="termsOfService" checkout="[[checkout]]"></lancie-terms-of-service>
+    <lancie-terms-of-service id="termsOfService">
+      <paper-button on-tap="closeDialog">Cancel</paper-button>
```
martijnjanssen: Can this be changed to a `dialog-dismiss`?

src/lancie-ticket-transfer/lancie-ticket-transfer-accept.html
```diff
@@ -56,16 +52,14 @@
       </div>
     </lancie-card>
 
-    <paper-dialog id="termsofservice" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
+    <lancie-dialog id="termsofservice" entry-animation="scale-up-animation" exit-animation="fade-out-animation" with-backdrop>
```
martijnjanssen: Unused attributes can be removed.

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,72 +1,177 @@
...
+
+This element is designed to give you a quick dialog with handy defaults,
+while at the same time being highly customizable.
+
+## Styling ##
+### Defaults ###
+To take advantage of the default styling you should use:
+`h2`            for the dialog header
+`.card-content` for the main content of the dialog
+`.buttons`      for the buttons at the bottom of the dialog
```
martijnjanssen: I'm not sure this is all correct, at some places in the code, `card-actions` is used. In the lancie-dialog code the buttons class is used. This has to be resolved, it all has to be the same.

I would choose for a rename to `dialog-actions`.

src/lancie-ticket-page/lancie-profile-dialog.html
```diff
@@ -37,20 +36,21 @@
...
       <h2>Fill in your profile</h2>
       <div class="card-content">
         <profile-edit-form id="form" user="{{user}}" on-response="checkSucceeded"></profile-edit-form>
       </div>
       <div class="card-actions">
```
martijnjanssen: This should be `dialog-actions` right? If you change this, the css class should also reflect this.

FrancisBehnen:
Todo:
 - [x] T&C checkout buttons
 - [ ] Set seatmap reserve seat button flush
 - [ ] Look into margin of lancie-activity-dialog combined with scrolling
 - [ ] Look into Martijn's small remarks
 - [ ] Remove lorem ipsum

src/lancie-activity/lancie-activity-dialog.html
```diff
@@ -60,55 +66,40 @@
...
-        background-color: black;
-        opacity: 0;
-        z-index: 10;
-        transition: opacity 6s;
-      }
-
       @media (max-width: 600px) {
-        .dialog-content {
+        .card-content {
           flex-direction: column;
```
FrancisBehnen: What did you have in mind? `min-width` on `lancie-activity-content`?

src/lancie-my-area/my-area-orders.html
FrancisBehnen: Nope, unfortunately in can't. This is because the button does not end up in the light-dom of the `lancie-dialog` if I'm not mistaken (@TimvdLippe). The button's parent `.dialog-actions` is, but the button itself isn't..

FrancisBehnen: If we merge this I will make an issue for this, and investigate at a later point if this is fixable without reverting to the old method.

scripts/activity.json
```diff
@@ -18,7 +18,7 @@
...
...na aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?",
```
TimvdLippe: Reminder :)

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,72 +1,177 @@
-<link rel="import" href="../../bower_components/paper-dialog/paper-dialog.html">
+<link rel="import" href="../../bower_components/polymer/polymer.html">
+<link rel="import" href="../../bower_components/iron-media-query/iron-media-query.html">
+
+
+<!--
+`lancie-dialog`
+Dialog for the lancie frontends
```
TimvdLippe: Good documentation, nice!

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,72 +1,177 @@
...
+      * Closes the lancie-dialog
+      * @return {void}
+      */
+      close: function() {
+        this.$.actualDialog.style.display = 'none';
+        this.$.backdrop.style.display = 'none';
+        this.$.backdrop.style.opacity = '0.0';
+      },
+
+      _viewChanged: function(param) {
```
TimvdLippe: `param` is unused

src/lancie-login/lancie-login-dialog.html
```diff
@@ -95,17 +65,6 @@
           this.fire('navigate', e.detail);
           this.$.dialog.close();
         }.bind(this));
-
-        window.document.body.appendChild(this.$.dialog);
```
TimvdLippe: :heart:

src/lancie-my-area/my-area-orders.html
TimvdLippe: Let's leave it like this. This is because it is inside two `<content>`.

src/lancie-ticket-page/lancie-ticket-basket.html
TimvdLippe: Nope, because it is inside two `<content>`

src/lancie-dialog/lancie-dialog.html
```diff
@@ -1,72 +1,177 @@
...
+      * Closes the lancie-dialog
+      * @return {void}
+      */
+      close: function() {
+        this.$.actualDialog.style.display = 'none';
+        this.$.backdrop.style.display = 'none';
+        this.$.backdrop.style.opacity = '0.0';
+      },
+
+      _viewChanged: function() {
```
martijnjanssen: [This](https://github.com/AreaFiftyLAN/lancie-frontend/pull/364/files#diff-86f53e86c659df83902a8d193c7d0f47R178) is what I am referring to.

FrancisBehnen: The use case is very small, but I actually sometimes resize my browser window while browsing a website. We can remove this as soon as native custom css properties are common.

FrancisBehnen:
Damn this PR, tested and reviewed and retested so many times, and still a major bug 😔

PR #369, opened 2017-01-09
Fix safari wrap

### LINKED ISSUE #368: Flex wrap does not work in Safari
If you give an element a `flex-[number]` class, `iron-flex-layout-classes` adds a `flex: [number];` css styling to that element. 

`flex: 1;` in css is shorthand for 
```
flex-grow: 1;
flex-shrink: 1;
flex-basis: 0%;
```
All browsers infer this correctly, except Safari. Safari needs a little help. The solution is to specify what it should actually infer. So replace `flex: 1;` with `flex: 1 1 0%` everywhere, except that the 0% doesn't work either.. so this has to be replaced again with the `min-width` of the element...

For reference:
- This is the bug report: https://bugs.webkit.org/show_bug.cgi?id=136041#c2
- I got my solution from here: http://stackoverflow.com/questions/25360526/row-wrap-in-flex-box-not-wrapping-in-safari/30792851#30792851
- The flex-layout developers don't see it as their problem: https://github.com/PolymerElements/iron-flex-layout/issues/35

FrancisBehnen:
Fixes #368

### DIFF
--- a/src/lancie-my-area/lancie-my-area.html
+++ b/src/lancie-my-area/lancie-my-area.html
@@ -12,12 +12,26 @@
     <style>
     :host {
       display: block;
+      --min-card-width: 300px;
     }
 
     .card {
       margin: 8px;
-      min-width: 300px;
+      min-width: var(--min-card-width);
     }
+
+    .flex-1 {
+      flex: 1 1 var(--min-card-width);
+    }
+
+    .flex-2 {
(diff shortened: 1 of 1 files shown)


FrancisBehnen:
Of course, we can also edit the `iron-flex-factors` module itself :p

FrancisBehnen:
@TimvdLippe, don't merge WIP PRs plz :p #378

PR #370, opened 2017-01-09
Enable ticketpage & my area

martijnjanssen:


MatthijsKok:
LGTM. Let's do this after backend is populated.

PR #388, opened 2017-01-10
Change ticket types for live deadline display

martijnjanssen:
Users can't see the ticket deadlines due to wrong matching.


MatthijsKok:
LGTM. Those are values in our live DB.

PR #391, opened 2017-01-11
Remove time zone usage

### LINKED ISSUE #390: Date is parsed as GMT time
This results in the ticket deadline being 1:00 instead of 00:00.

martijnjanssen:
Fixes #390 

Use UTC timezone since we are not working with a timezone. We don't have to compensate for time zone changes, so usage of the default user timezone is not needed.

### DIFF
--- a/src/lancie-ticket-page/lancie-ticket-item.html
+++ b/src/lancie-ticket-page/lancie-ticket-item.html
@@ -210,7 +210,7 @@
         }
         var deadlineType = this.deadlineDisplay[ticket.ticketType];
         if (deadlineType === 'deadline') {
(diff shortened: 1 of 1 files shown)


FrancisBehnen:
This is not the right fix imo. I think the api should provide the timezone information and the frontend should parse it according to that.

martijnjanssen:
But how soon will that be fixed? I would opt for merging this and chaning when the api provides timezone information.

skamoen:
Agreed. As it's a display issue, this can be easily fixed this way, and improved later on. It's not that this is _wrong_

skamoen:
An API fix is far from trivial (timezones are the worst), so this is by far the best solution for now.

PR #392, opened 2017-01-12
Check in .htaccess

### LINKED ISSUE #347: Check in htaccess and throw 404 on file lookups
Lets check in the htaccess file of the server into fit and also let it throw 404s on file lookups instead of redirecting to index.html

TimvdLippe:
File requests now return 404 when requested. This solves some nasty infinite loading loops when a file was missing in the build.

I have already updated the `.htaccess` on the server, this is simply checking in the resource.

Do we want this file also in the build?

Fixes #347


MatthijsKok:
I think we do. It is configurable after all. For example when we want to support additional file types.

MatthijsKok:
If this is already on the server and working, im merging this change.

PR #395, opened 2017-01-15
Dynamic close listener

FrancisBehnen:
Instead of responding to every tap event and figuring out whether it was on an element with the `dialog-dismiss` attribute this adds a `tap` `eventListener` to all elements with the `dialog-dismiss` attribute.

### DIFF
--- a/src/lancie-dialog/lancie-dialog.html
+++ b/src/lancie-dialog/lancie-dialog.html
@@ -147,8 +147,8 @@ simply give it the `dialog-dismiss` attribute.
         var host = this;
         var dismissEls = this.querySelectorAll('[dialog-dismiss]');
         for (var i = 0; i < dismissEls.length; i++) {
-          var dismissEl = dismissEls[i];
-          dismissEl.addEventListener('tap', function() { host.close(); });
+          dismissEls[i].addEventListener('tap',
+            function() { host.close(); });
         }
       },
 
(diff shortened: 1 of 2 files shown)


FrancisBehnen:
`dialog-dismiss` doesn't work on the `Close` button of the `<lancie-terms-of-service>` in [`<my-area-orders>`](https://github.com/AreaFiftyLAN/lancie-frontend/blob/69acb4c629551dfd4517c1386a60af1374f0ab69/src/lancie-my-area/my-area-orders.html#L77). This has probably to do with that the button is not defined directly in the `lancie-dialog`. Since this is an exceptional use case, I still consider this PR merge ready.

src/lancie-dialog/lancie-dialog.html
```diff
@@ -138,24 +138,17 @@
...
-          if (target.hasAttribute && (target.hasAttribute('dialog-dismiss') )) {
-            this.close();
-            event.stopPropagation();
-            break;
-          }
+      addDismissListeners: function() {
+        var host = this;
+        var dismissEls = this.querySelectorAll('[dialog-dismiss]');
+        for (var i = 0; i < dismissEls.length; i++) {
+          var dismissEl = dismissEls[i];
```
TimvdLippe: Unnecessary, shorter without the element

TimvdLippe:
@FrancisBehnen Yes that does not work as you do `querySelectorAll` which only sees lightdom. Since that button is passed into the slot, it is in the lightdom of `my-area-orders` and therefore not accessible by the `dialog`.

FrancisBehnen:
@TimvdLippe, are we going to use this. Or do you prefer sticking to the old method? (Which is a modification of the method used in `paper-dialog-behavior`)

TimvdLippe:
@FrancisBehnen Yes this is fine by me. Just don't have the time to actually test it. That's up to you and @MatthijsKok

FrancisBehnen:
@MatthijsKok I've already click-tested myself, so merge whenever you want.

PR #401, opened 2017-01-21
Fix sponsor logos flex wrapping in safari (and hopefully safari mobile)

FrancisBehnen:
If the previous fixes in `my-area` fixed it for safari mobile this works for safari mobile too.


martijnjanssen:
I think this is going to work, thanks!

PR #413, opened 2017-01-28
Serve with docker

martijnjanssen:
This is all copied from [this PR](https://github.com/AreaFiftyLAN/lancie-admin/pull/11) which is now live.


FrancisBehnen:
Works on the lancie-admin, and you kept the copy in this time ;) So guess it's ok

PR #414, opened 2017-01-28
Seatmap speedup

### LINKED ISSUE #397: SeatMap Speedup
Seatmap is very slow right now because we load in a dialog for every seat.
We need to make one dialog which loads in user info dynamically.

Needs #364

FrancisBehnen:
Fixes #397 
Needs #364

### DIFF
--- a/src/lancie-seatmap/lancie-seatmap.html
+++ b/src/lancie-seatmap/lancie-seatmap.html
@@ -41,7 +41,7 @@ This code may only be used under the BSD style license found at https://github.c
     <lancie-ajax auto id="ajaxGetTeamTickets" refurl="tickets/teammembers" on-response="onGetTeamTickets"></lancie-ajax>
     <lancie-ajax id="ajaxReserveSeat" method="POST" on-response="onReserveSeat"></lancie-ajax>
 
-    <paper-input label="Search user" value="{{filterUser}}"></paper-input>
+    <paper-input label="Search user" value="[[filterUser]]"></paper-input>
     <div class="layout horizontal">
       <div>
        <img src="/images-optimized/seatmap/stage.png">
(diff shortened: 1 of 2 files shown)


FrancisBehnen:
Can someone verify that this actually works? I'm not sure if it's doing what it's supposed to do...

martijnjanssen:
The speedup is indeed working, I will look at the code once the lancie-dialog is merged.

MatthijsKok:
This removes about 400ms of JS render time on my machine.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -62,6 +67,28 @@
...
+        <div hidden$="{{_checkExists(tickets)}}">
+          You need at least one ticket to reserve a seat!
+        </div>
+      </div>
+      <div class="dialog-actions">
+        <paper-button dialog-dismiss>Close</paper-button>
+        <paper-button on-tap="reserveSeat" hidden$="{{!_checkExists(tickets)}}">Reserve seat</paper-button>
+      </div>
+    </lancie-dialog>
+
```
martijnjanssen: All the properties in here are read only, so the bindings can change to one-way bindings.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -115,9 +146,42 @@
...
+        } else {
+          this.reservationError = true;
+          if (response.request.status === 409) {
+            this.$.error.setError('Seat is already taken.');
+          } else {
+            this.$.error.setError('Something went wrong, please refresh and try again.');
+          }
+        }
+      },
+      open: function(listenerEvent) {
```
martijnjanssen: I would rename the `listenerEvent` to just `e` or `event`, this is how we do it in the rest of the code.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -115,9 +146,42 @@
...
+        }
+        this.$.ajaxReserveSeat.params = {
+          'ticketId': this.$.listbox.selectedItem.ticket
+        };
+        this.$.ajaxReserveSeat.refurl = 'seats/' + this.seat.seatGroup + '/' + this.seat.seatNumber;
+        this.$.ajaxReserveSeat.generateRequest();
+        this.$.seatDialog.close();
+      },
+      onReserveSeat: function(e, response) {
+        if (response.succeeded) {
```
martijnjanssen: `response` should be renamed to `request`. This is something I have done incorrectly in the past.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -79,6 +106,7 @@
             return [];
           }
         },
+        seat: Object,
```
martijnjanssen: I would rename this to something along the lines of: `seatToReserve`. Explains a bit why there is a random single seat in here.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -115,9 +146,42 @@
...
+        this.$.ajaxReserveSeat.refurl = 'seats/' + this.seatToReserve.seatGroup + '/' + this.seatToReserve.seatNumber;
+        this.$.ajaxReserveSeat.generateRequest();
+        this.$.seatDialog.close();
+      },
+      onReserveSeat: function(e, request) {
+        if (request.succeeded) {
+          this.$.listbox.selected = undefined;
+          this.refreshSeats();
+        } else {
+          this.reservationError = true;
```
TimvdLippe: Is this unused?

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -90,6 +118,9 @@
           notify: true
         }
       },
+      listeners: {
+        'openDialog': 'open'
```
TimvdLippe: I dont like the naming `openDialog`. The event is actually a `requestSeatReserve` or something.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -115,9 +146,42 @@
           this.$.getAllTicketsErrorToast.show();
         }
       },
+      reserveSeat: function() {
+        this.$.error.clear();
+        if (!this.$.listbox.selectedItem) {
+          this.$.error.setError('Please select a user.');
+          return;
+        }
+        this.$.ajaxReserveSeat.params = {
```
TimvdLippe: Hm I thought this should be body instead of params. Probably a quirk in the API.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -62,6 +67,28 @@
...
+            </template>
+          </paper-listbox>
+        </div>
+        <div hidden$="[[_checkExists(tickets)]]">
+          You need at least one ticket to reserve a seat!
+        </div>
+      </div>
+      <div class="dialog-actions">
+        <paper-button dialog-dismiss>Close</paper-button>
+        <paper-button on-tap="reserveSeat" hidden$="[[!_checkExists(tickets)]]">Reserve seat</paper-button>
```
TimvdLippe: We could remove this and do an `on-iron-selected` on the listbox. We can keep it in this PR for now, maybe a thought for later

src/lancie-seatmap/lancie-seatmap.html
skamoen: I believe this was changed, as it was weird indeed. I'll check

skamoen: See https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/seats/controller/SeatRestController.java#L115

FrancisBehnen:
So we're not gonna speedup it further? I thought @martijnjanssen and @TimvdLippe said it was still too slow.

PR #415, opened 2017-01-29
Change to api subdomain

martijnjanssen:


src/lancie-login/lancie-login-form.html
```diff
@@ -37,7 +37,7 @@
 
     <iron-ajax
       id="getCurrentUser"
-      url="/api/v1/users/current"
+      url="https://areafiftylan.nl/v1/users/current"
```
praseodym: `api.`

martijnjanssen:
It still is, since the part before and including the `api` part will be replaced by the localhost url.

martijnjanssen:
Looks like the `api/v1/` is being successfully replaced when it passes through the stream. However, it does not work on the `lancie-app.html` and `lancie-homepage.html`. It seems like these are not being streamed for some reason. In these files, the replace action is not successful. Any idea how to fix this?

TimvdLippe:
I think instead you should replace on the stream at https://github.com/AreaFiftyLAN/lancie-frontend/blob/master/gulp-tasks/project.js#L57

martijnjanssen:
Tested and live after [this PR](https://github.com/AreaFiftyLAN/lancie-admin/pull/15) was merged

PR #416, opened 2017-01-29
Fix lancie timeline

FrancisBehnen:
Well, that was easier then expected :p

This does not fix the issue about the timeline since that was about rewriting the thing.

I want this pr to be about the technical side of the timeline. We'll create a separate PR to update the actual content.


MatthijsKok:
- [ ] Update the content of the timeline.
- [ ] Rebase off master instead of seatmap-speedup.

FrancisBehnen:
Content update in #417

martijnjanssen:
This does not work a 100% yet. The thing that is missing is the initially selected tab!

FrancisBehnen:
That is a bug that's been in the toolbar already for months. Lets fix it in a separate PR.

martijnjanssen:
LGTM.

PR #417, opened 2017-01-29
Update timeline content

FrancisBehnen:
Needs #416


FrancisBehnen:
Times match our own schedule, but our own schedule is not realistic. Blocked until we have a realistic schedule.

PR #422, opened 2017-01-31
Added hack for lancie-dialog errors in edge/ie

### LINKED ISSUE #421: Edge (and IE) doesn't support forEach() on NodeList
In #364 we introduced a `.forEach()` loop on the object returned by a `querySelectorAll()` call. Apparently Edge doesn't support this method on this object type (`NodeList`).

skamoen:
Fixes #421


FrancisBehnen:
@MatthijsKok Can you confirm that the dismiss functionality indeed works in Edge? E.g. the dialog dismisses if you click "close" on the "Remove a team member" dialog.

PR #427, opened 2017-02-13
Add Cooler Master as sponsor to CS:GO and Overwatch

MatthijsKok:


martijnjanssen:
LGTM.

martijnjanssen:
Do we actually want to list the prices of the prizes? I would prefer to leave it out and let the users search for it themselves.

svenpopping:
LGTM

PR #429, opened 2017-02-14
Update seatmap

FrancisBehnen:
Update seat map to the layout of this year


FrancisBehnen:
I'm not sure what to do with the inline styling. Is this okay or do we want to do this another way? Does someone have a suggestion to do this nicer?

TimvdLippe:
Please put the inline styles as normal CSS in the component

martijnjanssen:
Why don't you just put the styling in the style section with the rest of the css?

FrancisBehnen:
But how? Give it all id's?

TimvdLippe:
You usually use classes for that. You could in theory also use an attribute selector, but just put a class on it

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -35,6 +35,27 @@
                   0 1px 5px 0 rgba(0, 0, 0, 0.12),
                   0 3px 1px -2px rgba(0, 0, 0, 0.2);
     }
+
+    .seats {
+      height: 60vh;
+    }
+
+    .lounge {
+      height:500px;
```
martijnjanssen: Please place a space after the `:`.

MatthijsKok:
Please add a few pixels of space between the images. Rest looks good

MatthijsKok:
LGTM

PR #437, opened 2017-02-16
Replace toggle function

martijnjanssen:
A toggle is not needed here, there is no reason for it. The toggle function also does not exist anymore, resulting in breaking code.

Please test this commit, since it is completely untested.


martijnjanssen:
The dialog is now unable to have scrolling content, this needs to be fixed.

PR #443, opened 2017-02-17
Allow saving profile edit with enter

### LINKED ISSUE #372: Profile edit form submit on enter doesn't change to non-edit view
When I submit the profile edit form in My Area, the regular edit buttons are still there and the form is still displayed. 

Submit on enter should trigger the same things as clicking the checkmark button

MatthijsKok:
Fixes #372

### DIFF
--- a/src/lancie-my-area/my-area-profile.html
+++ b/src/lancie-my-area/my-area-profile.html
@@ -92,7 +92,7 @@
             </div>
           </div>
         </div>
-        <profile-edit-form id="profileForm" hidden$="[[!editingProfile]]" user={{user}}></profile-edit-form>
+        <profile-edit-form id="profileForm" hidden$="[[!editingProfile]]" user={{user}} on-profile-form-submit="submitEdit"></profile-edit-form>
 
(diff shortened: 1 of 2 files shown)


FrancisBehnen:
Don't merge yet, I have the feeling something is wrong here. Let me investigate tonight/tomorrow

FrancisBehnen:
As I expected, this makes it impossible to change gender without using a mouse to click. I liked that it was possible to change everything in your profile without needing your mouse. I think we should keep that feature.

MatthijsKok:
What are you talking about? I can easily traverse the entire form with enter and tab.

martijnjanssen:
It works in both chrome and firefox. If something does not work, please state: WHERE, WHEN, HOW. If you don't do this, we can't fix it.

FrancisBehnen:
Ah, nvm. I used enter to bring up the gender dropdown, but you can do that with up/down arrow. Then it's fine.

martijnjanssen:
LGTM.

PR #447, opened 2017-02-21
Add rules and schedules to official tournaments.

MatthijsKok:


MatthijsKok:
Francis' changes LGTM

PR #448, opened 2017-02-21
Fix My Area Seatcard

### LINKED ISSUE #445: Seat section in my-area not displaying reserved seats
When a user has reserved seats, they do not show up in the 'Seats' section in my-area. This section should have the reserved seats listed for the logged in user.

MatthijsKok:
Fixes #445

### DIFF
--- a/src/lancie-my-area/my-area-seat.html
+++ b/src/lancie-my-area/my-area-seat.html
@@ -23,8 +23,8 @@
     }
     </style>
 
-    <!--<lancie-ajax id="ajaxGetAllSeats" refurl="seats" on-response="onGetAllSeats"></lancie-ajax>-->
-    <!--<lancie-ajax auto id="ajaxGetSeat" refurl="users/current/seat" on-response="onGetSeat"></lancie-ajax>-->
+    <lancie-ajax id="ajaxGetAllSeats" refurl="seats" on-response="onGetAllSeats"></lancie-ajax>
+    <lancie-ajax auto id="ajaxGetSeat" refurl="users/current/seat" on-response="onGetSeat"></lancie-ajax>
     <lancie-ajax id="ajaxGetTeamTickets" refurl="tickets/teammembers" on-response="onGetTeamTickets"></lancie-ajax>
 
     <paper-card heading="Seat" elevation="1" animated-shadow="true">
--- a/src/lancie-my-area/my-area-seat.html
+++ b/src/lancie-my-area/my-area-seat.html
@@ -23,21 +23,9 @@
     }
     </style>
 
-    <lancie-ajax
-      auto
-      id="ajaxGetSeat"
-      refurl="users/current/seat"
-      on-response="onGetSeat"></lancie-ajax>
-
-    <lancie-ajax
-      id="ajaxGetAllSeats"
-      refurl="seats"
-      on-response="onGetAllSeats"></lancie-ajax>
-
-    <lancie-ajax
-      id="ajaxGetTeamTickets"
-      refurl="tickets/teammembers"
-      on-response="onGetTeamTickets"></lancie-ajax>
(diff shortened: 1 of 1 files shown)


martijnjanssen:
Could you add error handling for the requests? There are `console.log()` statements in there, but can be replaced by the global toast.

martijnjanssen:
Do we want to show our users a specific error message or a `Server unavailable`-like error message? An example of this can be found in https://github.com/AreaFiftyLAN/lancie-frontend/blob/938f090994bcaf51f364ac1bdced5321dea6ada0/src/lancie-ticket-page/lancie-ticket-item.html#L169.

src/lancie-my-area/my-area-seat.html
```diff
@@ -71,37 +83,37 @@
...
-            for (var j = 0; j < group.length; j++) {
-              if (group[j].ticket && this.teamMembers.includes(group[j].ticket.owner.profile.displayName)) {
-                seats[seats.length] = group[j];
+          var seatmapObject = request.response.seatmap;
+          var seatmapKeys = Object.keys(seatmapObject);
+          for (var i = 0; i < seatmapKeys.length; i++) {
+              var seatGroup = seatmapObject[seatmapKeys[i]];
+              for (var j = 0; j < seatGroup.length; j++) {
+                  if (seatGroup[j].ticket && this.teamMembers.includes(seatGroup[j].ticket.owner.profile.displayName)) {
+                      seats.push(seatGroup[j]);
```
martijnjanssen: I believe this entire action can be done with a `.filter(...)` call. This code is way too complex for what it is doing.

src/lancie-my-area/my-area-seat.html
```diff
@@ -23,9 +23,21 @@
...
+
+    <lancie-ajax
+      id="ajaxGetAllSeats"
+      refurl="seats"
+      on-response="onGetAllSeats"></lancie-ajax>
+
+    <lancie-ajax
+      id="ajaxGetTeamTickets"
+      refurl="tickets/teammembers"
+      on-response="onGetTeamTickets"></lancie-ajax>
```
martijnjanssen: Are the one-line requests really that difficult to read?

src/lancie-my-area/my-area-seat.html
```diff
@@ -23,9 +23,21 @@
     }
     </style>
 
-    <!--<lancie-ajax id="ajaxGetAllSeats" refurl="seats" on-response="onGetAllSeats"></lancie-ajax>-->
-    <!--<lancie-ajax auto id="ajaxGetSeat" refurl="users/current/seat" on-response="onGetSeat"></lancie-ajax>-->
-    <lancie-ajax id="ajaxGetTeamTickets" refurl="tickets/teammembers" on-response="onGetTeamTickets"></lancie-ajax>
+    <lancie-ajax
+      auto
```
martijnjanssen: If you do it like this, please place the auto behind the `lancie-ajax` and the closing bracket on the new line like this:
```html
<lancie-ajax auto
    id="ajaxGetSeat"
    refurl="users/current/seat"
    on-response="onGetSeat"
></lancie-ajax>
```

src/lancie-my-area/my-area-seat.html
MatthijsKok: I can't really make this code any cleaner with `.filter(...)`, since then the filtered array would have to be concatenated to `seats`, which is a very slow operation.

martijnjanssen: How do you mean? We only need to filter out the seats that belong to teammembers, this can be done by iterating overall seats, just in a different manner than with some for loops. This would increase the readability of the code.

FrancisBehnen: We want this feature live now. Can someone make an issue for this?

PR #450, opened 2017-02-25
Check fragments in polymer.json

TimvdLippe:
We can run this task with `yarn run build ensure-lazy-fragments`. Shall we also add this to Travis?

### DIFF
--- a/gulp-tasks/ensure-lazy-fragments.js
+++ b/gulp-tasks/ensure-lazy-fragments.js
@@ -1,0 +1,30 @@
+'use strict';
+
+const Analyzer = require('polymer-analyzer').Analyzer;
+const FSUrlLoader = require('polymer-analyzer/lib/url-loader/fs-url-loader').FSUrlLoader;
+const PackageUrlResolver = require('polymer-analyzer/lib/url-loader/package-url-resolver').PackageUrlResolver;
+const HtmlCustomElementReferenceScanner = require('polymer-analyzer/lib/html/html-element-reference-scanner').HtmlCustomElementReferenceScanner;
+const dom5 = require('dom5');
+const fragments = require('../polymer.json').fragments;
+
+let analyzer = new Analyzer({
+  urlLoader: new FSUrlLoader('./'),
+  urlResolver: new PackageUrlResolver(),
+  scanners: new Map([['html', [new HtmlCustomElementReferenceScanner()]]])
+});
+
+module.exports = () => new Promise((resolve, reject) => {
+  analyzer.analyze('src/lancie-content.html')
+  .then((document) => {
+    for (const element of document.getByKind('element-reference')) {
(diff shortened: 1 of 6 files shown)


martijnjanssen:
Yes please, would be great to add this to travis!

martijnjanssen:
Looks good!

gulp-tasks/ensure-lazy-fragments.js
```diff
@@ -0,0 +1,30 @@
...
+  scanners: new Map([['html', [new HtmlCustomElementReferenceScanner()]]])
+});
+
+module.exports = () => new Promise((resolve, reject) => {
+  analyzer.analyze('src/lancie-content.html')
+  .then((document) => {
+    for (const element of document.getByKind('element-reference')) {
+      const astNode = element.astNode;
+      if (astNode.parentNode.tagName === 'iron-lazy-pages') {
+        const childLocation = dom5.getAttribute(astNode, 'data-path');
```
tony19: For a more general solution, might be a good idea to read `<iron-lazy-pages>.attrForSelected` here instead of using `data-path` (i.e., get the attr name from `astNode.parentNode` and pass it to `dom5.getAttribute`).

gulp-tasks/ensure-lazy-fragments.js
```diff
@@ -0,0 +1,30 @@
...
+        if (!fragments.find(e => e === `src/${childLocation}`)) {
+          reject(`Could not find fragment "src/${childLocation}" in polymer.json!`);
```
tony19: The loop continues even after `reject`. http://stackoverflow.com/a/32536083/6277151

Return here for efficiency?

gulp-tasks/ensure-lazy-fragments.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const Analyzer = require('polymer-analyzer').Analyzer;
+const FSUrlLoader = require('polymer-analyzer/lib/url-loader/fs-url-loader').FSUrlLoader;
+const PackageUrlResolver = require('polymer-analyzer/lib/url-loader/package-url-resolver').PackageUrlResolver;
+const HtmlCustomElementReferenceScanner = require('polymer-analyzer/lib/html/html-element-reference-scanner').HtmlCustomElementReferenceScanner;
+const dom5 = require('dom5');
+const fragments = require('../polymer.json').fragments;
+
+let analyzer = new Analyzer({
```
tony19: const?

gulp-tasks/ensure-lazy-fragments.js
tony19: Oh, I see now that `data-path` means something else, so never mind.

TimvdLippe:
Thanks @tony19 for the review! I have updated the code accordingly :smile:

PR #451, opened 2017-02-27
Fix old firefox

FrancisBehnen:


skamoen:
Tested in Chrome, Chrome Canary, Firefox 50, Firefox 51, Firefox 48, IE11 and Edge. Good enough for me.

PR #452, opened 2017-02-27
Change configuration to lancie-admin

martijnjanssen:
We need to make this change to be able to switch to docker in production.


skamoen:
If you say so

PR #464, opened 2017-02-28
Add missing prizes

FrancisBehnen:


skamoen:
Prizes without sponsors still show the TBA placeholder. I'd replace this with something else.

PR #477, opened 2017-03-07
Revealing header on mobile

martijnjanssen:
The header now condenses and the links are always uppercase.


martijnjanssen:
We'll rework the entire composition of the pages with a different header and drawer layout on the homepage and the other pages.

PR #490, opened 2017-06-04
2.0 preview

TimvdLippe:
This makes the frontend homepage work in Polymer 2. There are some things I found out:

- Activity dialog is broken
- Gold email input does not have a 2.0 version, can't we just use `<input type="email">`?
- There are some elements which have horrible code. Not sure if these are salvagable or that we just rewrite them.

To test this branch:

```
rm -rf bower_components/
bower install
yarn run serve
```

You can check the amount of linter warnings with `polymer lint`. We still have some related to `iron-lazy-pages`, I will need to take a look there.


elarb:
Seems like the login dialog isn't behaving as intended as well: 

image

MatthijsKok:
Thanks a lot for your review! Could you provide some information about your browser / OS?

TimvdLippe:
Yes all dialogs are broken. Thus we must decide what to fix and what to redesign/reimplement

elarb:
I'm using Chrome 58 on Windows 10 @MatthijsKok

FrancisBehnen:
Safari can't find `webcomponents-lite.min.js` (`https://localhost:5100/bower_components/webcomponentsjs/webcomponents-lite.min.js`)

TimvdLippe:
We should eagerly load the loader instead

FrancisBehnen:
No idea what's going on here @TimvdLippe 

If I click on "Log in" in the master branch, it works of course:
screen shot 2017-07-01 at 14 32 08

But on this branch the `validate` function is not defined..
screen shot 2017-07-01 at 14 35 41

`validate` should be a function of `iron-form`, I checked, and the function is still there in the element.

TimvdLippe:
@FrancisBehnen Per the README of https://github.com/PolymerElements/iron-form it should be updated to the following:

```
<iron-form>
  <form>
  </form>
</iron-form>
```

FrancisBehnen:
Right, that worked. On to the next error 🤦‍♀️ 

screen shot 2017-07-01 at 15 52 14


```
if (typeof this.headers === 'object') {
        console.log(this.headers);
        for (header in this.headers) {
          console.log(header);
          console.log(this.headers[header]);
          headers[header] = this.headers[header].toString();
        }
      }

```
`iron-ajax` tries to convert the `X-Auth-Token` to a String, but since there's no token.

TimvdLippe:
Then either insert the token or put it to an empty string?

FrancisBehnen:
But why does it have a token anyway? On the master the `this.headers` object is just empty.
screen shot 2017-07-01 at 16 03 58

TimvdLippe:
I don't know :joy: Maybe look up in the stacktrace to see where it is set?

Op za 1 jul. 2017 om 16:04 schreef FrancisBehnen <notifications@github.com>:

FrancisBehnen:
Something is going wrong with making the call to the api. Whether I try to log in with right or wrong credentials, I always get a `request.succeeded` and an empty token in the `on-response` handler in the `lancie-login-form` (https://github.com/AreaFiftyLAN/lancie-login-form/blob/2.0-preview/lancie-login-form.html#L90).

PR #491, opened 2017-06-04
Update dev dependencies

TimvdLippe:
Updates polymer-build to the stable version, feels safer. I also ported the configuration from https://github.com/PolymerElements/generator-polymer-init-custom-build/blob/master/generators/app/gulpfile.js


skamoen:
Let's merge this into the 2.0-preview branch and work from there

PR #493, opened 2017-09-01
Update to 2.0, fix styling and flow

martijnjanssen:
The drawer now always hides on mobile and goes away when a page is opened. On desktop is stays open untill the homepage opens, then it is closed.


skamoen:
Now the menu in the top right corner is gone. For clarity, I'm not sure that's an improvement.

martijnjanssen:
We'll fix this another way, this is not really the best solution.

PR #504, opened 2017-09-16
Tabbed ticket page

### LINKED ISSUE #409: Dialogs in order process perceived as annoying
We currently have 3 intercepts in the Order proces

1. Login
2. Profile check
3. Terms and conditions

Each dialog requires the user to click the same button again, which is perceived as annoying and slow. We should streamline this better by chaining them together, without the need to click the same button in between.

martijnjanssen:
Fixes #409

A lot of the current ticket ordering system is perceived as annoying, this is mostly due to all the dialogs during the order process and lack of predictability.


martijnjanssen:
This is true, and is fixed in #511

martijnjanssen:
I think this is mostly it, the only thing I need to do is check the flows:
- Anonymous order and couple it
- Anonymous order and couple it after registering a new user
- Creating an order with an existing user

Any more testing flows? I can't really think of other cases that are possible.

martijnjanssen:
BUG: Logging in and submitting the form with the enter key returns the user to the first tab for some reason

Fastjur:
To add to that bug: Pressing the `Next` key does not actually log the user in, and simply returns to the first tab

martijnjanssen:
I'm fairly sure that the issues have all been resolved, please test all possible flows by clicking manually. Please only report issues relevant to this PR, other issues can be added in a normal issue on the repo.

skamoen:
Minor issues:
- I always get a "profile updated" toast, even if I don't change anything. Might be confusing
- I kinda expect a "login" button under the form. Clicking "next" with my credentials filled in feels counter-intuitive. Adding a button "login" which does the same as "next" solves this.

Registering is awkward. You have to click the "next" button to submit the form. It then resets the form... and nothing happens. You do get a register email, but that contains a default confirmation link, which takes you to my-area instead of continuing the order. 
When going to the tickets page, it did remember my order (localstorage?) and I can login. All ok from here. 

The profile form for new users is fully red. it's not wrong, but feels "punishy" while you didn't have the chance to fill it in yet. Maybe call validate opon clicking next? Filling them out doesn't change their color upon unfocus.

What I'm expecting when registering:
Fill out form, click "Register" somewhere. This removes the form, and shows a notification that I should check my email for verification. The link in the email should directly continue the order proces (we have support for this in de api!) and drop you on the profile section.

martijnjanssen:
- The profile update toast can best be resolved in the `profile-edit-form` by not even submitting the profile to the server when changes are made. This is the easiest and best solution in my opinion.
- The next button should even be a login/register button for the specific actions in the form, which is even better.

For registering, the registering form should be fixed first before I am going to fix this here in this branch.

After the register form is fixed, I'll look at the registering and the registering flow including the order on a different device.

martijnjanssen:
I have now added the order coupling when registering and added a check to remove the first time validation penalty for new users without a profile.

index.html
```diff
@@ -200,7 +200,7 @@
 
           <lancie-login-check data-requires-user data-route="my-area" data-path="src/lancie-my-area/lancie-my-area.html">
             <template>
-              <lancie-my-area data-requires-user></lancie-my-area>
+              <lancie-my-area user="[[user]]"></lancie-my-area>
```
TimvdLippe: :party_parrot:

index.html
```diff
@@ -260,7 +260,7 @@
   <script>
     'use strict';
     (() => {
-      const requireUsers = document.querySelectorAll('data-requires-user');
+      const requireUsers = document.querySelectorAll('[data-requires-user]');
```
TimvdLippe: Me thinking how I could write this: :man_facepalming:

src/lancie-layout/lancie-layout.html
```diff
@@ -0,0 +1,155 @@
+<link rel="import" href="../../bower_components/app-layout/app-header-layout/app-header-layout.html">
```
TimvdLippe: I think this file should have been removed already

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
 
-    <lancie-ajax id="ajaxAvailable" refurl="tickets/available" on-lancie-ajax="onAvailableTickets"></lancie-ajax>
+    <lancie-ajax auto-fire skip-token id="ajaxAvailable" refurl="tickets/available" on-lancie-ajax="onAvailableTickets"></lancie-ajax>
     <lancie-ajax id="ajaxCoupleOrder" refurl="orders/[[basket.id]]/assign" method="POST" on-lancie-ajax="onCoupleOrder"></lancie-ajax>
-    <lancie-ajax auto-fire refurl="users/current/orders/open" method="GET" on-lancie-ajax="onOpenOrder"></lancie-ajax>
+    <lancie-ajax id="ajaxCheckout" refurl="orders/[[basket.id]]/checkout" method="POST" on-lancie-ajax="onCheckout"></lancie-ajax>
+    <lancie-ajax auto-fire refurl="users/current/orders/open" on-lancie-ajax="onOpenOrder"></lancie-ajax>
+
+    <div class="paper-material" elevation="1" on-next="next">
+      <paper-tabs id="step" selected="{{selected}}">
```
TimvdLippe: All tabs are disabled, so maybe just use an `iron-selector` and do the (easy) styling yourself. See the homepage for an example in my PR

TimvdLippe: Oh and since you are not using the tabs anyways, the `selected` can be one-way

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
+        <paper-tab disabled>Tickets</paper-tab>
+        <paper-tab disabled>Login</paper-tab>
+        <paper-tab disabled>Profile</paper-tab>
+        <paper-tab disabled>Terms &amp; Conditions</paper-tab>
+      </paper-tabs>
+      <paper-progress min="-1" value="[[selected]]" max="3"></paper-progress>
```
TimvdLippe: -1?

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
+          <lancie-ticket-basket class="flex-2" basket="[[basket]]" user="{{user}}"></lancie-ticket-basket>
+        </div>
 
-      <lancie-tickets class="flex-3" tickets="[[availableTickets]]" basket="[[basket]]"></lancie-tickets>
+        <lancie-login id="login" order-id="[[basket.id]]" registering="{{registering}}" on-logged-in="onLogin"></lancie-login>
 
-      <lancie-ticket-basket class="flex-2" basket="[[basket]]" user="{{user}}"></lancie-ticket-basket>
+        <div>
+          <h2>Fill in your profile</h2>
+          <profile-edit-form id="profileForm" user="{{user}}" on-profile-submitted="profileConfirmed"></profile-edit-form>
```
TimvdLippe: Did we confirm all other user values were updated as well? Can't remember if I broke this or not.

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
+        </div>
 
+        <div>
+          <h2>Terms and conditions</h2>
+          <lancie-terms-of-service-text></lancie-terms-of-service-text>
+        </div>
+
+      </iron-pages>
+      <div class="buttons layout horizontal end-justified">
```
TimvdLippe: I really dislike these layout classes. Rather use declarative CSS instead. But this is something we might have to address everywhere else as well.

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
+    properties: {
+      user: Object,
+      basket: {
+        type: Object,
+        value: {},
+        notify: true,
+      },
+      availableTickets: {
+        type: Object,
+        value: [],
```
TimvdLippe: The type is Object. Shouldn't it be this?

```js
function() {
  return {};
}
```

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
+        this.$.profileForm.validateAndSubmit();
+      } else if (this.selected === 3) {
+        this.$.ajaxCheckout.generateRequest();
+      } else {
+        this.selectTab(0);
+      }
+    },
+
+    selectTab: function(tab) {
+      if (tab === 2 && !!this.user.profile.displayName) {
```
TimvdLippe: `tab === 2` #magicnumber If you want to keep the logic like this, add a comment to explain what it actually means

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
+
+    storeStorageOrder: function (e) {
+      if (!this.basket.id) {
+        this.basket = e.detail;
+      }
+    },
+
+    onLogin: function() {
+      // Required to wait after the user has propogated down into the dom
+      this.async(this.next);
```
TimvdLippe: Ah I see why you opened that other issue. Yeah this is annoying.

src/lancie-ticket-page/lancie-ticket-page.html
```diff
@@ -5,156 +5,308 @@
...
-            this.$.ajaxCoupleOrder.generateRequest();
-          } else if (basket.status === 'ANONYMOUS' && !this.$.orderStorage.orderStored()) {
-            this.$.orderStorage.storeOrderId(basket.id);
-          }
-        },
-
-      });
-    })();
+  (function () {
+  'use strict';
```
TimvdLippe: There are various magic numbers used which correspond to tabs. Let's declare an enum above to make the code more readable.

```js
const TABS = {
  TICKETS: 0,
  LOGIN: 1,
  PROFILE: 2,
  CONDITIONS: 3
};
```

src/lancie-ticket-page/lancie-ticket-page.html
martijnjanssen: Yeah we dont need the tabs after all, since we are using the progress bar to indicate where you are in the ordering system. I'll change it to some divs, they don't even need different styling 😂

src/lancie-ticket-page/lancie-ticket-page.html
martijnjanssen: Yup, since the iron-pages start at 0 and go to 3. So at the first page we want to show the progress under the first step, at the last page, the entire progress bar is filled.

src/lancie-ticket-page/lancie-ticket-page.html
martijnjanssen: I'm fairly sure it was only broken at the `login-check` level, since this profile form always has the correct data from the user when at this point in the ticket page.

src/lancie-ticket-page/lancie-ticket-page.html
martijnjanssen: Correct, this is also a dependency we can get rid of everywhere that makes this all less magic-y.

src/lancie-ticket-page/lancie-ticket-page.html
martijnjanssen: Yup, otherwise they all all get the same value when editing one of them ;), we had this issue before.

src/lancie-ticket-page/lancie-ticket-page.html
martijnjanssen: After some code inspection, this function was not even needed after all 🎉

martijnjanssen:
We are now fully done 🎉

Fastjur:
# Good weather

1. Register while ordering:  
There is something funky going on after registering and trying to login. The enter key does not redirect the user. Only when clicking on the login button does it redirect to the tickets page again. I have attempted this multiple times:

* https://giant.gfycat.com/IndolentTemptingCurlew.webm
* https://giant.gfycat.com/SparseUntidyCuttlefish.webm

Worth noting here is the warning `listener method 'onLogin' not defined`

2. Login while ordering:  
Experienced similar issues with login during ticket ordering process. Also had to refresh once because I could not agree to the terms and agreement.
https://giant.gfycat.com/FriendlyDigitalHairstreakbutterfly.webm

3. Order, when already logged in:  
*Works fine and dandy!*

# Bad weather

1. Register while ordering, try to login without validating:  
*Works as expected*

2. Cancelled payment (and tried spam adding some tickets):  
*Works as expected*  
Interesting to note, the enter key now magically did work when trying to log in.

3. Order status Open:  
*Works as expected*  
Note, when clicking on the info button next to said panding order in my area, the ticket type in the dialog states `Owned`, which may be confusing. 

# Conclusion
This is working quite well and is a big improvement. There are some weird issues however, that do not seem to always be reproducible.

PR #506, opened 2017-09-19
Load FAQ from API

MatthijsKok:


src/lancie-faq/lancie-faq.html
```diff
@@ -19,21 +19,22 @@
...
+    handle-as="json"
     on-response="onResponse"
-    on-error="onError"></iron-ajax>
+  ></lancie-ajax>
 
-  <marked-element markdown="[[text]]">
-    <div class="markdown-html"></div>
-  </marked-element>
+  <template is="dom-repeat" items="[[faq]]" as="faqitem">
+    <h3>[[faqitem.question]]</h3>
```
TimvdLippe: Are we certain both h1 and h2 are used before this? Dont use h3 just for styling. Use the proper heading and then style with css.

MatthijsKok: the markdown document uses `### Question` which according the markdown spec directly translates to a `<h3>` element

martijnjanssen (approved):

martijnjanssen:
All done, only needs adminpanel implementation.

ArdyZ:
@MatthijsKok and @martijnjanssen can you have a look at this pull request and what should be done before it can be merged?

martijnjanssen:
I think it's a matter of checking whether the API does what is required from an implementation standpoint here. If that is fulfilled, and testing it works, then it *should* be fine. I don't have that much knowledge about these old PR's anymore though.

PR #508, opened 2017-09-19
Load Tournaments from API

MatthijsKok:


src/lancie-home-page/lancie-home-page.html
```diff
@@ -116,7 +116,7 @@
           tournament, we will give you the opportunity to organize your own. More information about how this works can
           be found below.</p>
 
-        <lancie-activities json="scripts/official-tournaments.json?ver=1.1"></lancie-activities>
+        <lancie-activities tournament-type="OFFICIAL"></lancie-activities>
```
TimvdLippe: Default is already official.

skamoen:
Blocked until we have CRUD functionality in lancie-admin

TimvdLippe:
I fear this PR is rather ancient and therefore the merge becomes rather nasty. Let's do this again, but from the most recent master. That should bypass most of the git issues and use the new ES6 notation :tada:

PR #511, opened 2017-09-21
Adapt new polymer 2.0 ES6 login check

martijnjanssen:


src/lancie-login/lancie-login-check.html
```diff
@@ -1,111 +1,94 @@
...
+            '_canAccess(user, nologin)'
+          ];
+        }
 
-      observers: [
-        '_canAccess(user, nologin)'
-      ],
+        ready() {
+          super.ready();
+          checks.push((newUser) => {
```
TimvdLippe: Lets do this in connectedcallback and remove it in disconnectedcallback

src/lancie-login/lancie-login-check.html
```diff
@@ -1,111 +1,94 @@
...
+            this.user = newUser;
+          });
+        }
 
-      ready: function() {
-        checks.push(function(newUser) {
-          this.user = newUser;
-        }.bind(this));
+        _canAccess(user, nologin) {
+          if (!this.stamped && (nologin || (user && user !== ''))) {
```
TimvdLippe: Add stamped as an argument to this observer if you depend on it

src/lancie-login/lancie-login-check.html
```diff
@@ -1,111 +1,94 @@
...
+            let template = this.querySelector('template');
```
TimvdLippe: I think the template is in the lightdom, so you must retrieve the content from the slot? That can be done with `slot.assignedNodes()`

src/lancie-login/lancie-login-check.html
```diff
@@ -1,111 +1,94 @@
...
+            let TemplateClass = Polymer.Templatize.templatize(template, this, {});
```
TimvdLippe: Should there be forwarding of databindings?

src/lancie-login/lancie-login-check.html
martijnjanssen: So set the user on connect and remove the user on the disconnect (set it to `{}`)?

src/lancie-login/lancie-login-check.html
martijnjanssen: The method does not work when only called with the template as an argument, and we were not sure what the 2nd and 3rd parameter do.

src/lancie-login/lancie-login-check.html
TimvdLippe: No, remove the lambda from the array

src/lancie-login/lancie-login-check.html
skamoen: I think @TimvdLippe  means this: 
>forwardHostProp(property, value): Called when a property referenced in the template changed on the template's host. As this library does not retain references to templates instanced by the user, it is the templatize owner's responsibility to forward host property changes into user-stamped instances. The instance.forwardHostProp(property, value) method on the generated class should be called to forward host properties into the template to prevent unnecessary property-changed notifications. Any properties referenced in the template that are not defined in instanceProps will be notified up to the template's host automatically.

I don't understand what they mean, but as it's currently working as intended, I don't think we need to specify it... Or is that too simple?

src/lancie-login/lancie-login-check.html
```diff
@@ -1,111 +1,88 @@
...
 
-      ready: function() {
-        checks.push(function(newUser) {
-          this.user = newUser;
-        }.bind(this));
+        _canAccess(user, nologin) {
+          if (nologin || (user && user !== '')) {
+            let template = this.$.slot.assignedNodes().find((element) => element.nodeName === 'TEMPLATE');
+            let TemplateClass = Polymer.Templatize.templatize(template, this, {});
+            var instance = new TemplateClass({user: user});
```
martijnjanssen: Is there any downside or difference by assigning the user immediately?

src/lancie-login/lancie-login-check.html
martijnjanssen: This is not outdated, even though GitHub marks it as such :-1: @TimvdLippe

src/lancie-login/lancie-login-check.html
TimvdLippe: If the user would change, that would not be updated. You can fix that by implementing the `forwardHostProp` which should call `_setPendingPropertyOrPath` on the template instance. See https://github.com/Polymer/polymer/blob/374d4079e215a8246e3981e2f0c58edc44a1239e/lib/utils/templatize.html#L109-L123

src/lancie-login/lancie-login-check.html
```diff
@@ -14,98 +13,87 @@
...
+          if (!instance && (nologin || (user && user !== ''))) {
+            let template = this.$.slot.assignedNodes().find((element) => element.nodeName === 'TEMPLATE');
+            let TemplateClass = Polymer.Templatize.templatize(template, this, {
+              forwardHostProp(prop, value) {
+                if(this.instance) {
+                  this.instance.forwardHostProp(prop, value);
+                }
+              },
+              instanceProps: {
+                user: true,
```
martijnjanssen: For some reason, even without this property, the changes in user made are reflected in the underlying data-bindings. For instance, when the `user` object in `lancie-content` is changed, the changes can be seen in `my-area` when inspecting, but the values on screen are not changed. I can't seem to figure out what is wrong here...

src/lancie-login/lancie-login-check.html
```diff
@@ -14,98 +13,87 @@
...
+          checks.splice(checks.indexOf(this.callbackFunction), 1);
+        }
 
-        // Add eventlistener to catch login dialog open request
-        this.addEventListener('login-dialog', function(e) {
-          if (!!e.detail.orderId) {
-            this.$.loginDialog.orderId = e.detail.orderId;
+        _canAccess(instance, user, nologin) {
+          if (!instance && (nologin || (user && user !== ''))) {
+            let template = this.$.slot.assignedNodes().find((element) => element.nodeName === 'TEMPLATE');
```
TimvdLippe: Should be `.toLowerCase()`: https://johnresig.com/blog/nodename-case-sensitivity/

PR #513, opened 2017-09-21
Change committee member icons to use position

Fastjur:
Changes the behaviour of the commissioner list.

Now the commissioner's position is used for determining the icon needed.
This needs to be taken into account when adding / editing the commission members in the admin page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "fs-extra": "^4.0.1",
     "glob": "^7.1.1",
     "gulp": "gulpjs/gulp#4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,6 +7,7 @@
     "fs-extra": "^4.0.1",
     "glob": "^7.1.1",
     "gulp": "gulpjs/gulp#4.0",
+    "gulp-babili": "^0.1.4",
```
TimvdLippe: Should be removed.

martijnjanssen: How did this even end up here? :joy:

PR #517, opened 2017-10-12
Fix icon imports for commission members

Fastjur:


Fastjur:
@skamoen @martijnjanssen 

Decide what you want. I'm fucking done with this one

elarb:
Instead of having to provide both the set and the icon name, I would propose only sending the icon name from the API, e.g. `icons:gavel` -> `gavel` and always use our own set with only icons we need.

src/lancie-commission/lancie-commission-member.html
```diff
@@ -1,12 +1,8 @@
 <link rel="import" href="../../bower_components/iron-flex-layout/iron-flex-layout-classes.html">
-<link rel="import" href="../../bower_components/iron-icon/iron-icon.html">
```
TimvdLippe: Um, should this be removed? We are still using iron-icon.

elarb: Good catch, I was getting a little too excited removing lines from the code 😄

elarb:
Can I have a confirmation before merging this?

TimvdLippe:
Does this work without modifying the API? Do we still want to change the API format? If so please file an issue on that repo

elarb:
The API is compatible

TimvdLippe:
It will break when we publish live though, so we have to take that into mind

@skamoen @martijnjanssen @MatthijsKok When we publish live, we have to remind ourselves to update these values to the ones in the testdatarunner.

PR #518, opened 2017-10-12
Fix broken event propagation in login-check

martijnjanssen:


martijnjanssen:
Logging in from a child of login-check doesn't work at all when these lines are commented. When a page does not require logging in, but does allow a user to log in from (for example the ticket page), the event will top at the login-check (which should not happen). This should be testable in the old ticket page, although it can also be tested in #504.

martijnjanssen:
Merged with approval of @TimvdLippe in Slack.

PR #521, opened 2017-10-23
Add mail subscription option

### LINKED ISSUE #520: Allow users to enter their email address to be notified when ticket sale starts
Implement a simple form to allow users to register on a mailing list for when ticketsale opens.

Fastjur:
Opening PR to be able to keep track for now.

Requires https://github.com/AreaFiftyLAN/lancie-api/pull/445

Fixes #520 

TODO:

- [x] Implement subscription in API

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,13 +7,12 @@
     "fs-extra": "^4.0.1",
     "glob": "^7.1.1",
     "gulp": "gulpjs/gulp#4.0",
-    "gulp-babel-minify": "^0.2.0",
     "gulp-html-minifier": "^0.1.8",
     "gulp-if": "^2.0.1",
-    "gulp-imagemin": "^3.3.0",
     "gulp-jshint": "^2.0.2",
     "gulp-replace": "^0.6.1",
     "gulp-style-modules": "^0.1.0",
+    "gulp-babel-minify": "^0.2.0",
     "jshint": "^2.9.1",
     "jshint-stylish": "^2.0.0",
     "merge-stream": "^1.0.0",
@@ -21,7 +20,8 @@
     "polymer-analyzer": "^2.0.0-alpha.28",
     "polymer-build": "^2.0.0",
     "polyserve": "0.20.0",
-    "superagent": "^3.2.1"
+    "superagent": "^3.2.1",
+    "gulp-imagemin": "^3.3.0"
   },
   "scripts": {
     "build": "node ./node_modules/gulp/bin/gulp.js",
--- a/package.json
+++ b/package.json
@@ -7,12 +7,13 @@
     "fs-extra": "^4.0.1",
     "glob": "^7.1.1",
     "gulp": "gulpjs/gulp#4.0",
+    "gulp-babel-minify": "^0.2.0",
     "gulp-html-minifier": "^0.1.8",
     "gulp-if": "^2.0.1",
+    "gulp-imagemin": "^3.3.0",
     "gulp-jshint": "^2.0.2",
     "gulp-replace": "^0.6.1",
     "gulp-style-modules": "^0.1.0",
-    "gulp-babel-minify": "^0.2.0",
     "jshint": "^2.9.1",
     "jshint-stylish": "^2.0.0",
     "merge-stream": "^1.0.0",
@@ -20,8 +21,7 @@
     "polymer-analyzer": "^2.0.0-alpha.28",
     "polymer-build": "^2.0.0",
     "polyserve": "0.20.0",
-    "superagent": "^3.2.1",
-    "gulp-imagemin": "^3.3.0"
+    "superagent": "^3.2.1"
   },
   "scripts": {
     "build": "node ./node_modules/gulp/bin/gulp.js",
--- a/src/lancie-home-page/lancie-home-page.html
+++ b/src/lancie-home-page/lancie-home-page.html
@@ -81,9 +81,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     <lancie-image-slider></lancie-image-slider>
 
-    <lancie-section type="primary" header="Stay up to date!">
+    <lancie-section type="primary" header="Updates">
       <div>
-        Enter your E-Mail here to be notified when the tickets are for sale!
+        <p>
+          Don't worry, we hate spam too. We will not share your details with anyone.<br/>
+          You will only receive emails relevant to the event, such as when
+          the ticket sales are opened.
+        </p>
         <lancie-subscribe-mail-form></lancie-subscribe-mail-form>
       </div>
     </lancie-section>
--- a/src/lancie-home-page/lancie-home-page.html
+++ b/src/lancie-home-page/lancie-home-page.html
@@ -19,6 +19,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../lancie-faq/lancie-faq.html">
 <link rel="import" href="../lancie-facebook/lancie-facebook.html">
 <link rel="import" href="lancie-image-slider.html">
+<link rel="import" href="lancie-subscribe-mail-form.html">
 
 <dom-module id="lancie-home-page">
   <template>
@@ -80,9 +81,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     <lancie-image-slider></lancie-image-slider>
 
-    <lancie-section type="primary" header="Partners &amp; Sponsors">
-      <lancie-sponsors></lancie-sponsors>
+    <lancie-section type="primary" header="Stay up to date!">
+      <div>
+        Enter your E-Mail here to be notified when the tickets are for sale!
+        <lancie-subscribe-mail-form></lancie-subscribe-mail-form>
+      </div>
     </lancie-section>
+
+    <!--<lancie-section type="primary" header="Partners &amp; Sponsors">
+      <lancie-sponsors></lancie-sponsors>
+    </lancie-section>-->
 
     <lancie-section type="secondary" header="Area FiftyLAN">
       <div class="intro-container">
--- a/src/lancie-home-page/lancie-subscribe-mail-form.html
+++ b/src/lancie-home-page/lancie-subscribe-mail-form.html
@@ -34,7 +34,7 @@
       <lancie-error id="mailFormError"></lancie-error>
       <!-- Can be replaced with lancie-alert later on -->
       <lancie-error id="mailFormInfo"></lancie-error>
-      <paper-input label="E-Mail" name="email" class="mail-input" value="{{email}}">
+      <paper-input label="E-Mail" name="email" class="mail-input" value="{{email}}" on-keydown="checkForEnter">
         <iron-icon icon="mail" slot="prefix"></iron-icon>
         <paper-icon-button slot="suffix" on-tap="clearInput" icon="clear" alt="clear" title="Clear"></paper-icon-button>
       </paper-input>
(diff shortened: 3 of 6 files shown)


martijnjanssen:
Sorry, but I don't fully agree on the dependency on AreaFiftyLAN/lancie-api#114. That issue is responsible for sending the mail, which is not the change here. This PR depends on an additional endpoint for adding a user in the "**Save the date**" mailing list, which should be created in the API. With that you could manually get the mail from the API through the adminpanel and still be able to send the mail in a regular mailing client, just as we did in previous years.

Fastjur:
We currently have a choice. I am using https://github.com/AreaFiftyLAN/lancie-frontend/pull/521/files#diff-5e006dbd67bda15d1a6943ee3aad5ff1R72 to retrieve the hibernate error messages. (Such as `Email should be valid`). This is very neat as we can define the error messages on the constraints themselves. However, when no user friendly error message is set, we get error messages back like this: image

package.json
```diff
@@ -7,21 +7,21 @@
...
     "gulp-style-modules": "^0.1.0",
-    "gulp-babel-minify": "^0.2.0",
     "jshint": "^2.9.1",
     "jshint-stylish": "^2.0.0",
     "merge-stream": "^1.0.0",
     "plylog": "^0.5.0",
     "polymer-analyzer": "^2.0.0-alpha.28",
     "polymer-build": "^2.0.0",
     "polyserve": "0.20.0",
-    "superagent": "^3.2.1",
```
martijnjanssen: What happened here? 👎

src/lancie-home-page/lancie-home-page.html
```diff
@@ -54,15 +55,26 @@
     </style>
     <lancie-image-slider></lancie-image-slider>
 
-    <lancie-section type="primary" header="Partners &amp; Sponsors">
-      <lancie-sponsors></lancie-sponsors>
+    <lancie-section type="primary" header="Updates">
+      <div>
+        <p>
+          Don't worry, we hate spam too. We will not share your details with anyone.<br/>
```
martijnjanssen: Wouldn't two `<p>` elements be just as easy?

src/lancie-home-page/lancie-home-page.html
```diff
@@ -54,15 +55,26 @@
...
+    <!--<lancie-section type="primary" header="Partners &amp; Sponsors">
+      <lancie-sponsors></lancie-sponsors>
+    </lancie-section>-->
+
     <lancie-section type="secondary" header="Area FiftyLAN">
       <div>
         <p>
           Area FiftyLAN is the hottest gaming event of the TU Delft. This spectacular
-          all-weekend event, from 3 till 5 March 2017, will be organized by the study
+          all-weekend event, from 2 till 4 March 2018, will be organized by the study
```
martijnjanssen: Do we really want to update this now? All texts still need an overhaul.

src/lancie-home-page/lancie-subscribe-mail-form.html
```diff
@@ -0,0 +1,84 @@
...
+        color: var(--primary-color);
+      }
+
+      [hidden] {
+        display: none !important;
+      }
+    </style>
+
+    <lancie-form id="mailForm" refurl="subscriptions" method="POST" class="mail-input" on-response="_onResponse">
+      <lancie-error id="mailFormError"></lancie-error>
```
martijnjanssen: You can just as easy just call it `info` or `error`, it is a contained component after all.

src/lancie-home-page/lancie-subscribe-mail-form.html
```diff
@@ -0,0 +1,84 @@
...
+  </template>
+
+  <script>
+    (function() {
+      'use strict';
+
+      Polymer({
+        is: 'lancie-subscribe-mail-form',
+        properties: {
+          email: {
```
martijnjanssen: A simple `email: String,` is sufficient here :)

src/lancie-home-page/lancie-subscribe-mail-form.html
```diff
@@ -0,0 +1,84 @@
...
+        },
+
+        _onResponse: function(event, request) {
+            console.log(event, request);
+          if (request.succeeded) {
+            this.clearInput();
+            this.$.mailFormError.clear();
+            this.$.mailFormInfo.setError("Thank you for signing up!");
+          } else {
+            this.$.mailFormError.setError(
```
martijnjanssen: For this purpose, I have built the so called "Refresh-toast", [find the event here](https://github.com/AreaFiftyLAN/lancie-frontend/blob/420f0e69eca1babf42af8a2fac05edce64af916d/src/lancie-ticket-page/lancie-ticket-basket.html#L135).

And the code here: [the eventHandler](https://github.com/AreaFiftyLAN/lancie-frontend/blob/master/src/lancie-content.html#L111-L113)

src/lancie-home-page/lancie-subscribe-mail-form.html
```diff
@@ -0,0 +1,84 @@
...
+          this.$.mailFormError.clear();
+          this.$.mailFormInfo.clear();
+        },
+
+        _onResponse: function(event, request) {
+            console.log(event, request);
+          if (request.succeeded) {
+            this.clearInput();
+            this.$.mailFormError.clear();
+            this.$.mailFormInfo.setError("Thank you for signing up!");
```
martijnjanssen: You could also make a toast out of this but this is also okay. You can make a toast event and the `lancie-content` handles showing the toast, just letting you know. 👍

src/lancie-home-page/lancie-subscribe-mail-form.html
```diff
@@ -0,0 +1,84 @@
...
+          this.$.mailForm.validateAndSubmit();
+          this.$.mailFormError.clear();
+          this.$.mailFormInfo.clear();
+        },
+
+        _onResponse: function(event, request) {
+            console.log(event, request);
+          if (request.succeeded) {
+            this.clearInput();
+            this.$.mailFormError.clear();
```
martijnjanssen: If you're clearing the contents when sending the request, you don't need to clear the `mailFormError` here again.

src/lancie-home-page/lancie-subscribe-mail-form.html
```diff
@@ -0,0 +1,84 @@
...
+            this.$.mailFormInfo.setError("Thank you for signing up!");
+          } else {
+            this.$.mailFormError.setError(
+              !!request.request.response.message ? request.request.response.message : 'Oops! Something has gone wrong. Please refresh and try again.'
+            );
+            this.$.mailFormInfo.clear();
```
martijnjanssen: Same here, you cleared it when sending the request, so no need to do this again.

package.json
Fastjur: This happens automatically when running `yarn`. It cleans up the package.json

src/lancie-home-page/lancie-home-page.html
Fastjur: Quite right. Just as easy. I don't see the difference

src/lancie-home-page/lancie-home-page.html
Fastjur: I'd like the most critical points to be up to date right now. We are gonna overhaul the text before the save the date milestone. But at least this will be correct.

src/lancie-home-page/lancie-subscribe-mail-form.html
Fastjur: Once again `Just as easy`

src/lancie-home-page/lancie-subscribe-mail-form.html
Fastjur: Ah yep. Chagned

src/lancie-home-page/lancie-subscribe-mail-form.html
Fastjur: Yes it is. Else the error does not disappear. Note that `clearInput` clears the fields, and the `.clear()` calls are for the error and info dialog

src/lancie-home-page/lancie-subscribe-mail-form.html
Fastjur: I thought about that, but especially on bigger screens the toats may be missed. I think it's nice to see a clearly visible info dialog here to ensure people know they have been subscribed.

src/lancie-home-page/lancie-subscribe-mail-form.html
Fastjur: The second link 404's here. And what does it do?

src/lancie-home-page/lancie-subscribe-mail-form.html
Fastjur: Scratch that. I get what you are saying.

martijnjanssen: You're doing the exact same clear action [here](https://github.com/AreaFiftyLAN/lancie-frontend/pull/521/files#diff-5e006dbd67bda15d1a6943ee3aad5ff1R60), which is just after sending the request (I would advise clearing first and then sending the request).

src/lancie-home-page/lancie-subscribe-mail-form.html
Fastjur: Changed :+1:

src/lancie-home-page/lancie-subscribe-mail-form.html
martijnjanssen: Ah, these lines were moved to the `lancie-app` but these are the lines, since we will be moving that code one [more time](https://www.youtube.com/watch?v=FGBhQbmPwH8):
```javascript
this.addEventListener('refreshToast', function() {
    this.showToast({text: 'Unable to connect to server. Try refreshing the page.'});
}.bind(this));
```

src/lancie-home-page/lancie-home-page.html
martijnjanssen: I don't really like the `<br/>` in the code, and it is [bad to use it for this purpose](https://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br)

Fastjur: To quote from there `If you want two items in different visual blocks, you probably want them in different logical blocks.`

Which is not what I want. I want a line break between those 2 lines, to prevent one word wrapping on bigger screens. I do not want them in 2 visually different blacks.

Fastjur:
@martijnjanssen What do we want to do with the text changes in this PR? They're some small critical points that I want updated ASAP, we need to rewrite all the texts anyway.

Fastjur:
@martijnjanssen Just to let you know, this has been rebased. If you are gonna change the `lancie-error` to `lancie-alert` still, you'll have to pull it in again

Fastjur:
Since the API part has been merged, any issues merging this one? @martijnjanssen

src/lancie-home-page/lancie-subscribe-mail-form.html
```diff
@@ -0,0 +1,86 @@
...
+        color: var(--primary-color);
+      }
+
+      [hidden] {
+        display: none !important;
+      }
+    </style>
+
+    <lancie-form id="mailForm" refurl="subscriptions" method="POST" class="mail-input" on-response="onResponse">
+      <lancie-error id="alert"></lancie-error>
```
TimvdLippe: Suspicion that this `id` clashes with `window.alert` in Edge.

Fastjur: Thank you. Will take a look at it 👍

Issue #488, opened 2017-05-07
Change username fields to email


martijnjanssen:
The API will start using email instead of username for the main user identifier after [this PR](https://github.com/AreaFiftyLAN/lancie-api/pull/406) is merged.

martijnjanssen:
There is a big change that the password reset is currently broken and `my-area-ticket-item` has several username occurences that have to be changed to email: https://github.com/AreaFiftyLAN/lancie-frontend/blob/master/src/lancie-my-area/my-area-ticket-item.html

martijnjanssen:
Currently, password reset is completely broken since the email field is commented... Welp...

skamoen:
Fixed on all places we encountered it. make specific issues if we missed one

PR #522, opened 2017-10-25
Change field from email input to paper-input

martijnjanssen:
Part of #488 

Gold-email-input doesn't really add much, so since the element is no longer supported, we are moving to the simpler paper-input.


Fastjur:
Works fine and dandy :+1:

PR #525, opened 2017-10-28
Remove lancie-content

TimvdLippe:
This was mainly a wrapper around iron-lazy-pages.
However, this generated weird bindings where content
was suddenly doing logic with toasts, storage, etc...
Since there is no functional usage of lancie-content
other than 1 more file and less lines per file,
we should remove this file.

This will also be relevant for extracting the home page content
into index.html which is following this PR


martijnjanssen:
What is wrong about the current implementation of the toasts? We could also be scattering toasts all throught the code, and have them not work when a dialog is active, which is the other option.

For storage, where would you otherwise move this logic? This is the top-most component where the user is used, so this makes sense to me.

TimvdLippe:
There is nothing wrong with the implementation, but simply the placement of the toast. Eventually this will be moved into `index.html` which:

1. Unifies all logic into the entrypoint
2. Makes sure we capture ALL events and don't miss one in a sibling accidently
3. Don't need to move the toast to `window.document.body`

Likewise for storage, that should be at the top level, which is `index.html`.

MatthijsKok:
I like moving storage and toasts into index.html, but I do not think that the page content should be there as well...

TimvdLippe:
@MatthijsKok Per concerns of file size? Yes that is an issue. We could fix it on build time instead, will take a look :+1:

PR #527, opened 2017-10-29
Add testsuite

elarb:
- Added an initial test runner in the same way Polymer is running their tests
- Added a small initial test-suite for lancie-contact
- Using `paper-input` now for the email input

Users of the site should be able to submit feedback, report bugs, etc., so it is important that the contact form works properly.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -13,16 +13,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   <head>
     <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
-    <title>Elements Test Runner</title>
+    <title>LANcie Test Runner</title>
     <meta charset="UTF-8">
-
-    <script>
-      WCT = {
-        mochaOptions: {
-          timeout: 60000,
-        },
-      };
-    </script>
 
     <script src="../bower_components/webcomponentsjs/webcomponents-lite.js"></script>
     <script src="../bower_components/web-component-tester/browser.js"></script>
@@ -34,6 +26,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       let suites = [
         'unit/lancie-contact/lancie-contact.html',
       ];
+
+      // Runner setup inspired by the Polymer library: https://github.com/Polymer/polymer/blob/master/test/runner.html
 
       // http://eddmann.com/posts/cartesian-product-in-javascript/
       function flatten(arr) {
--- a/test/index.html
+++ b/test/index.html
@@ -30,36 +30,25 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Runner setup inspired by the Polymer library: https://github.com/Polymer/polymer/blob/master/test/runner.html
 
       // http://eddmann.com/posts/cartesian-product-in-javascript/
-      function flatten(arr) {
-        return [].concat.apply([], arr);
-      }
+      const flatten = (arr) => [].concat.apply([], arr);
 
-      function product(sets) {
-        return sets.reduce(function(acc, set) {
-          return flatten(acc.map(function(x) {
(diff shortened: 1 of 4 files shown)


test/unit/lancie-contact/lancie-contact.html
```diff
@@ -0,0 +1,61 @@
...
+      lancieContact.message = new Array(10).join(message);
+    });
+
+    test('instantiation works', function() {
+      assert.equal(lancieContact.is, 'lancie-contact');
+    });
+
+    test('request fired', function(done) {
+      this.timeout(500);
+      lancieContact.addEventListener('lancie-ajax', () => {
```
martijnjanssen: Is this even going to work? Since the api won't be available in travis.

test/index.html
```diff
@@ -16,16 +16,82 @@
     <title>Elements Test Runner</title>
     <meta charset="UTF-8">
 
+    <script>
+      WCT = {
+        mochaOptions: {
+          timeout: 60000,
```
TimvdLippe: This is too long for a regular test suite. I would vouch for keeping the default of 10 seconds

test/index.html
```diff
@@ -16,16 +16,82 @@
...
+        'unit/lancie-contact/lancie-contact.html',
+      ];
+
+      // http://eddmann.com/posts/cartesian-product-in-javascript/
+      function flatten(arr) {
+        return [].concat.apply([], arr);
+      }
 
+      function product(sets) {
+        return sets.reduce(function(acc, set) {
```
TimvdLippe: All functions can be lambdas, as we are not testing on IE

test/index.html
```diff
@@ -16,16 +16,82 @@
...
+      }
+
+      function addUrlOption(previous, next) {
+        previous = previous || '';
+        next = next || '';
+        return previous + (previous ? '&' : '') + next;
+      }
+
+      // test shadowdom/custom elements polyfills together
+      // prefering both if possible.
```
TimvdLippe: I suspect most of this logic is based on some other code? If so, please add a link like "inspired from ..."

test/unit/lancie-contact/lancie-contact.html
TimvdLippe: We could add some assertion here as well

elarb: @martijnjanssen Thought behind it was that the request would eventually be made given that the inputs are valid.
@TimvdLippe Yep, will add those!

test/index.html
elarb: Good point.

test/index.html
elarb: Yes, the biggest part is from the test runner Polymer is using, since I really liked how they set it up. Will add a link, thanks for reminding me :D

test/index.html
elarb: 👍

elarb:
I think that this is ready to be merged

test/unit/lancie-contact/lancie-contact.html
TimvdLippe: I think we still need this assertion?

test/unit/lancie-contact/lancie-contact.html
```diff
@@ -0,0 +1,61 @@
...
+        done();
+      });
+
+      lancieContact.shadowRoot.querySelector('paper-button').click();
+    });
+
+    //TODO
```
TimvdLippe: What needs to be done still?

elarb: Its not so clear what to assert there, and since the test depends on time and doesn't seem to be repeatable, I am considering removing it.

About the todo, the test is failing. Not sure why though. Will have to inspect `lancie-contact` validation a bit further.

TimvdLippe:
Rebased this PR as the build was failing. Should be fixed now.

PR #532, opened 2017-10-30
Preload home page content for fast first paint

TimvdLippe:
An initial full load can now be generated and is later
inserted into the slot of home-page.

Note that I had to remove the image slider to make this all work. Since this was scheduled for removal anyways, that shouldn't be a problem.

We should remove `<lancie-section>` as well, as we can just do `:nth-child` and use `<h2>` directly.


src/lancie-home-page/lancie-home-page.html
```diff
@@ -53,47 +47,15 @@
       }
     </style>
 
+    <slot></slot>
```
martijnjanssen: Above that `lancie-ajax`? Is this to prevent lancie-ajax loading first?

TimvdLippe: The issue with non-shadow dom supported browsers is that the elements are moved in the dom. Then `:nth-child` takes into account the `lancie-ajax`, e.g. flipping the colors.

martijnjanssen: Wouldn't a `:nth-child` with `:not(lancie-ajax)` work?

martijnjanssen:
The styling is completely missing in the content that is loaded before the lancie-home. Font family is the most obvious one.

TimvdLippe:
What do you mean with "Style completely missing"? I have a full screen paint with content that remains styled as is

martijnjanssen:
image
This is what happens in both chrome and firefox, a different font and letter size and spacing.

TimvdLippe:
Oh, I thought it was "completely missing". Probably missed several CSS styles, let's see.

index.html
```diff
@@ -178,7 +251,7 @@
           <paper-icon-button class="menu-button" icon="menu" drawer-toggle></paper-icon-button>
 
           <a href="/">
-            <img height="32" width="200" src="images-optimized/lancie/logo_without_date.png" />
+            <img class="logo" height="32" width="200" src="images-optimized/lancie/logo_without_date.png" />
```
elarb: Original image is 2143x349. Maybe we should resize that one?

martijnjanssen:
Is this completely finished? 😛

TimvdLippe:
Um yes, if it works ;)

src/lancie-home-page/lancie-home-page.html
TimvdLippe: No, it just counts the children sadly :(

Fastjur:
So what is the status here?

Should I test this or are there some known issues still?

TimvdLippe:
I am holding this off for now, we can do this change later.

Op ma 4 dec. 2017 om 14:42 schreef Jurriaan Den Toonder <
notifications@github.com>:

Fastjur:
Alright :+1:

TimvdLippe:
Have to think about this, but this is not the solution that is good enough. Therefore closing this PR.

PR #533, opened 2017-10-31
Resize the logo image

elarb:
Small changes, but it makes a difference:

Live:
image

Now:
image


TimvdLippe:
We automatically resize images in a gulp task. I think you need to update it there

PR #535, opened 2017-11-14
Re-add eventlistener to the index

martijnjanssen:
The navigate eventListener was previously removed by index (I assume), which resulted in the registration not working properly anymore and also non functioning navigation buttons and actions in other places.

### DIFF
--- a/index.html
+++ b/index.html
@@ -282,6 +282,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       document.addEventListener('toast', (e) => {
         showToast(e.detail);
       });
+      document.addEventListener('navigate', (e) => {
+        appLocation.path = e.detail.page;
+      });
       document.addEventListener('refreshToast', (e) => {
         showToast({text: 'Unable to connect to server. Try refreshing the page.'});
       });


martijnjanssen:
It does work, the only thing I was unsure about is whether this is the best way to set the page. I looked into the documentation of `app-location` and `app-route` briefly, and to me this was the most logical way of doing this, not completely sure about this though.

TimvdLippe:
Hm. How did we fix that previously? I don't recall having this listener somewhere

martijnjanssen:
We did have one, I think it was lost after moving some of the files since it was still right [here](https://github.com/AreaFiftyLAN/lancie-frontend/commit/3abeeb562033508eee1d3ec8f23314a603ba85b8#diff-64da152a18afecfd4c46f82c3340e958R239). Would it really be useful to re-add the async action in this PR as well?

TimvdLippe:
Oh right, then this is a valid solution. I simply forgot to migrate this listener. Async is not needed I think with Polymer 2, as we prevent an endless loop with the bag of properties.

martijnjanssen:
I'm merging this PR, the code was already there in previous versions but went missing somewhere along the way.

PR #550, opened 2017-12-16
Switch to Polymer CLI build

skamoen:
This PR migrates our build process to the Polymer CLI build process, instead of our custom build file. This simplifies the build process a lot, and allows for new possibilities such as [Differential Serving](https://github.com/Polymer/prpl-server-node)

Polymer.json now contains the build configuration. The custom project.js is removed. The string-replace task to replace all url's to areafiftylan.nl is now a separate task which we run on the builld files


package.json
```diff
@@ -24,7 +24,7 @@
     "gulp-imagemin": "^3.3.0"
   },
   "scripts": {
-    "build": "node ./node_modules/gulp/bin/gulp.js",
+    "build": "node ./node_modules/gulp/bin/gulp.js ensure-lazy-fragments && polymer build && node ./node_modules/gulp/bin/gulp.js replace-api",
```
TimvdLippe: `ensure-lazy-fragments` is unnecessary here. You already added it to the `default` task.

polymer.json
```diff
@@ -5,9 +5,14 @@
   "entrypoint": "index.html",
   "shell": "src/index-imports.html",
   "sources": [
+   "src/**/*",
```
TimvdLippe: Hm that's weird. Is this really necessary? :open_mouth:

package.json
TimvdLippe: Also we are now depending on the version of the polymer-cli. Should we maybe add this to the devdependencies and call it from there? That ensures we are all using the same version.

polymer.json
skamoen: >>>An optional array of globs matching your application source files. If left out, defaults to all files in your project src/ directory. You’ll need to set this if you store your source files in other locations.

We're not leaving it out, so I figured it had to be added. Also works without, so you're right.

skamoen:
Fixed Travis and processed initial feedback. Ready for a proper review!

martijnjanssen:
Looks like we are missing the linter in the build process now, let's get that back in here.

Dockerfile
```diff
@@ -5,4 +5,4 @@ RUN docker/dockerbuild.sh
 
 FROM nginx:mainline
 COPY docker/nginx_server.conf /etc/nginx/conf.d/default.conf
-COPY --from=builder /src/build /srv/www
\ No newline at end of file
+COPY --from=builder /src/build/fallback /srv/www
```
TimvdLippe: Don't we want `modern` here, as that loads ES6?

docker/dockerbuild.sh
```diff
@@ -3,11 +3,12 @@ set -euxo pipefail
 
 # Install programs required for build
 npm install -g yarn bower
```
TimvdLippe: Let's install `yarn` with the official instructions (https://yarnpkg.com/lang/en/docs/install/) and then do `yarn global add bower polymer-cli`

package.json
```diff
@@ -1,30 +1,23 @@
...
     "jshint-stylish": "^2.0.0",
-    "merge-stream": "^1.0.0",
     "plylog": "^0.5.0",
-    "polymer-analyzer": "^2.0.0-alpha.28",
-    "polymer-build": "^2.0.0",
-    "polyserve": "0.20.0",
-    "superagent": "^3.2.1",
-    "gulp-imagemin": "^3.3.0"
+    "polymer-analyzer": "^2.7.0",
+    "polymer-build": "^2.1.1",
```
TimvdLippe: We no  longer rely on this package.

package.json
```diff
@@ -1,30 +1,23 @@
...
+    "polyserve": "0.23.0",
+    "superagent": "^3.2.1"
   },
   "scripts": {
-    "build": "node ./node_modules/gulp/bin/gulp.js",
+    "build": "polymer build && node ./node_modules/gulp/bin/gulp.js replace-api",
```
TimvdLippe: We are implicitly relying on the version of `polymer` here. Not sure what the best approach is, but we should at least mention which version of the Polymer CLI we are relying on in the README.

Dockerfile
skamoen: #551 Handles differential serving. For this PR, only one build will be served so I went for the fallback build to be one the safe side.

docker/dockerbuild.sh
skamoen: As discussed in Slack, this will increase image size by a lot, without getting much in return.

docker/dockerbuild.sh
```diff
@@ -3,11 +3,12 @@ set -euxo pipefail
...
+npm install -g polymer-cli --unsafe-perm
 
 # Fetch dependencies
 yarn
 bower --allow-root install
 
 # Compress images and build
 yarn run build optimize-images
```
martijnjanssen: We can also remove this line and add `optimize-images` to [this line here](https://github.com/AreaFiftyLAN/lancie-frontend/pull/550/files#diff-b9e12334e9eafd8341a6107dd98510c9R88) right?

Fastjur: That would make sense

skamoen: This causes the build to run twice now, as the syntax of `yarn build` changed.

PR #551, opened 2017-12-16
Switch to PRPL Node.js server

skamoen:
This PR will migrate us from a nginx docker container to a prpl-server, capable of differential serving based on browser capabilities. 

Depends on #550 
## Testing
_Warning: This connects to the actual API. Live data!_
### Local
```
npm install -g prpl-server
yarn run build
prpl-server --root ./build --config polymer.json
```
Browse to [127.0.0.1:8080](http://127.0.0.1:8080)
### Docker
```
docker build -t lancie-frontend:latest .
docker run -p 80:80 lancie-frontend:latest
```
Browse to [localhost](localhost)


index.html
```diff
@@ -22,7 +22,7 @@
   <meta property="og:description" content="Area FiftyLAN is the hottest gaming event of Delft. For the third year in a row over 200 gamers will gather and play their favourite on- and off-line games together for three days (and nights)!" />
   <meta property="og:type" content="website" />
   <meta property="og:url" content="https://areafiftylan.nl/" />
-  <meta property="og:image" content="https://areafiftylan.nl/images-optimized/lancie/logo_without_date_with_background.png" />
+  <meta property="og:image" content="images-optimized/lancie/logo_without_date_with_background.png" />
```
skamoen: Not sure about this one, and didn't know how to test. If it needs to be absolute, it should be changed to `https://areafiftylan.nl/modern/images-optimized/(...)`

index.html
```diff
@@ -145,6 +145,7 @@
     }
   </style>
 
+  <base href="/">
```
TimvdLippe: I don't think we want to do this,  as this will screw up some of our URLs. Could you explain why this is necessary?

polymer.json
```diff
@@ -9,8 +9,8 @@
    "scripts/*"
   ],
   "builds": [
-    {"name":"fallback", "preset": "es5-bundled" },
-    {"name": "modern", "preset": "es6-unbundled", "browserCapabilities": ["es2015", "push"]}
+    {"name":"fallback", "preset": "es5-bundled", "basePath": true},
```
TimvdLippe: Ah, reading https://www.polymer-project.org/2.0/docs/tools/polymer-json#builds about the `basePath`, which makes sense. This means that older browsers (e.g. IE) get the ES5 version and modern browsers get ES6?

skamoen: Yes, the prpl-server determines the browser capabilities, and serves from a build folder based on that. This is also the reason we need the `<base>` tag, as this whole setup relies on it. 

You can test this following the instructions, and then going to `localhost` from different browsers. You'll see that Chrome loads from `modern` while Firefox loads from `fallback`

The URL changes are required for this setup. I'm not sure if there's a way of excluding static assets from the build paths, but it'll be come a much more complicated config if we want this.

index.html
skamoen: See comment below

polymer.json
```diff
@@ -9,8 +9,8 @@
...
+    {"name": "modern", "preset": "es6-unbundled", "browserCapabilities": ["es2015", "push"], "basePath": true}
```
TimvdLippe: I would remove `push` here, as it limits Edge and Firefox. I would rather serve ES6 to these browsers than just Chrome

TimvdLippe:
Looking at the follow network data, it appears we are not serving HTTP2.
image
Is this fixed by our reverse proxy on the CH server?

index.html
Fastjur: Meta tags should be able to resolve relative paths right?

gulpfile.js
```diff
@@ -63,6 +63,10 @@ gulp.task('replace-api', () => {
   return gulp.src(['build/**/*']).pipe(gulpif(/\.html$/, gulpreplace('/api/v1', 'https://api.areafiftylan.nl/api/v1'))).pipe(gulp.dest('build'))
 });
 
+gulp.task('replace-api', () => {
+  return gulp.src(['build/**/*']).pipe(gulpif(/\.html$/, gulpreplace('/api/v1', 'https://api.areafiftylan.nl/api/v1'))).pipe(gulp.dest('build'))
+});
```
TimvdLippe: This seems to be duplicate.

skamoen:
Build is still failing. I'm not sure I'll have time today to look into it

docker/dockerbuild.sh
```diff
@@ -9,4 +9,6 @@ npm install -g polymer-cli --unsafe-perm
 yarn
 bower --allow-root install
 
-yarn build
+# Compress images and build
+yarn run optimize-images
```
TimvdLippe: This should be `yarn run build optimize-images`

martijnjanssen: We can remove this line, take a look at the `package.json`, the build task runs checks (linter, optimizes images and checks fragments), builds and executes the post build scripts.

PR #556, opened 2017-12-19
Log out button

Jipje:
Work in progress.

Issues:

`<lancie-logout-button hidden$="[[!user]]"></lancie-logout-button>`  is not sufficient to hide the logout button when a user is not logged in.

Also we need to find a way to refresh the "cache" or local storage after logging out. Logging out brings you to the homepage but returning to the My Area page still shows the users information. Using ```window.location.replace("/");``` instead of a `navigate` event seemed to work...

### DIFF
--- a/index.html
+++ b/index.html
@@ -104,7 +104,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       display: flex;
     }
 
-    .navigation-tabs a {
+    .navigation-tabs a, .navigation-tabs lancie-logout-button {
       text-decoration: none;
       color: var(--secondary-color);
       font-size: 16px;
@@ -186,6 +186,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
               <a data-route="tickets" href="/tickets">Tickets</a>
               <a data-route="my-area" href="/my-area">My Area</a>
               <a data-route="seatmap" href="/seatmap">Seatmap</a>
+              <lancie-logout-button hidden$="[[!user]]"></lancie-logout-button>
             </iron-selector>
           </nav>
         </app-toolbar>
--- a/index.html
+++ b/index.html
@@ -186,7 +186,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
               <a data-route="tickets" href="/tickets">Tickets</a>
               <a data-route="my-area" href="/my-area">My Area</a>
               <a data-route="seatmap" href="/seatmap">Seatmap</a>
-              <lancie-logout-button hidden$="[[!user]]"></lancie-logout-button>
             </iron-selector>
           </nav>
         </app-toolbar>
--- a/src/lancie-login/lancie-logout-button.html
+++ b/src/lancie-login/lancie-logout-button.html
@@ -1,0 +1,83 @@
+<link rel="import" href="../../bower_components/polymer/polymer.html">
+<link rel="import" href="../../bower_components/lancie-ajax/lancie-ajax.html">
+<link rel="import" href="../lancie-storage/lancie-auth-storage.html">
+
+<dom-module id="lancie-logout-button">
+<template>
+  <style>
+    :host {
+      display: block;
+      height: 100%;
+      min-height: 100%;
+    }
+
+    :host [hidden] {
+        display: none !important;
+    }
+    
+    a {
+      text-decoration: none;
+      color: var(--secondary-color);
+      /*font-size: 16px;
+      font-weight: 400;
+      line-height: 24px;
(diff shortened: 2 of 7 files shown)


Fastjur:
So. The issue here is that the `iron-lazy-pages` loads the pages when clicked, but they are not removed after logging out. A `window.location.replace("/")` call works because the full page is then reloaded (and the `iron-lazy-pages` need to be reloaded again (causing the `login-check` to fail and thus showing a login).

I'm tagging @TimvdLippe  as he's the author of `iron-lazy-pages`. I know there is a way to force the inputs etc. to reload always, but I was unable to find a way to force them to be 'unloaded' or something similar.

src/lancie-storage/lancie-auth-storage.html
```diff
@@ -58,6 +58,20 @@
...
+        this.$.form.clearToken();
+      },
+      // TODO include these functions in their respective repositories.
+      // Feel free to use the following code for testing purposes
+      //
+      // clearToken: function() {
+      //  this.auth = {};
+      //  this.$.ajax.injectToken('');
+      //  this.$.form.injectToken('');
+      // },
```
Fastjur: This PR works as intended after we fix these issues

index.html
```diff
@@ -186,6 +186,7 @@
               <a data-route="tickets" href="/tickets">Tickets</a>
               <a data-route="my-area" href="/my-area">My Area</a>
               <a data-route="seatmap" href="/seatmap">Seatmap</a>
+              <lancie-logout-button hidden$="[[!user]]"></lancie-logout-button>
```
TimvdLippe: Is the CSS style not applied? In that case, you have to do `[hidden] { display: none !important; }`

src/lancie-login/lancie-logout-button.html
```diff
@@ -0,0 +1,83 @@
...
+      min-height: 48px;
+      height: inherit;
+      display: flex;
+      align-items: center;*/
+    }
+  </style>
+
+  <lancie-ajax id="ajaxLogout" refurl="logout" method="POST" on-lancie-ajax="onLogoutResponse"></lancie-ajax>
+
+  <a href="#" on-tap="tryLogout">Logout</a>
```
TimvdLippe: Shouldn't this be a `<button>` instead?

src/lancie-login/lancie-logout-button.html
```diff
@@ -0,0 +1,83 @@
...
+
+</template>
+<script>
+  (function () {
+  'use strict';
+
+  const authStorage = document.querySelector('lancie-auth-storage');
```
TimvdLippe: This is brittle, as we do not know whether this remains. In general: do not query outside from your element, only going down.

src/lancie-login/lancie-logout-button.html
```diff
@@ -0,0 +1,83 @@
...
+
+  class LancieLogoutButton extends Polymer.Element {
+    static get is() {
+      return 'lancie-logout-button';
+    }
+
+    static get properties() {
```
TimvdLippe: I think you can omit this if this element has no properties.

martijnjanssen:
The easiest solution to all of this is removing all the functionality we added and just calling the logout endpoint and refreshing the page. Doing it this way, the page takes care of all the expired tokens and not giving access to users not logged in. Way easier, sorry for only realizing this now.

Fastjur:
Yes that would be easiest. However I recall that most adblock plugins prevent a page from refreshing, and some browsers too. I think we would end up not having this work in 10/30% (educated guess) of the users.

martijnjanssen:
But as discussed, there isn't any perfect solution to circumvent this, we we're just going with the solution I proposed right?

Fastjur:
Yes after the discussion in Slack I think that is the best approach!

index.html
elarb: This actually won't work at all since it won't bind to anything.

elarb:
I moved the button to My Area:

logout-button

elarb:
What is the state of this PR? Do we want to look more into where to place the button?

Fastjur:
Wouldn't we want to move this to the header? In the top right?

It should be possible to hide it when not logged in right.

elarb:
@Fastjur Yes, I think we should do that. I took a look at it earlier today but something else caught my attention, but will look at it again later

Fastjur:
Alright thank you!

Op ma 12 feb. 2018 om 18:16 schreef e_lar <notifications@github.com>:

TimvdLippe:
Closing this, as rebasing becomes very nasty against current master. Let's retry this with a fresh PR.

PR #559, opened 2018-01-02
Fix pages not rendering in edge

### LINKED ISSUE #554: Polyfill issue in Edge
In Edge, the content doesn't load at all, and throws the following error:

>Object doesn't support property or method 'getElementById'

@elarb Did some research, found that
>"this error usually occurs when an html element id has the same id as some variable in the javascript function"

@martijnjanssen Suggested it's likely somewhere in lancie-ajax and/or lancie-form

Fastjur:
Closes #554 

Signed-off-by: Jurriaan Den Toonder <jur.den.toonder@gmail.com>


skamoen:
Doesn't fix anything

PR #562, opened 2018-01-07
CSS tweaks and fixes

martijnjanssen:
This PR actually fixes the colors in firefox ~and edge~. The problem was caused by the CSS leaking to the other elements due to not properly using shadow dom in edge and firefox, R126 fixes this.

The other issue we had was the dialog for the order in my area floating to the left, which was caused by a leftover CSS statement from previous dialog implementation.

Lastly, I removed some redundant CSS due to the fact that the lancie-error element now handles all the styling and colors. I should have spotted this in #521, but failed to.

### DIFF
--- a/index.html
+++ b/index.html
@@ -56,14 +56,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       color: #333;
       background-color: #eeecec;
       margin: 0;
-      --lancie-primary: #1a2b43;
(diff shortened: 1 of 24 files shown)


martijnjanssen:
I just assumed that the issue in firefox and edge were identical, but they aren't. At least this fixes one of them.

skamoen:
It improves things anyway, merging.

PR #564, opened 2018-01-10
Fix slash error for webcomponents

### LINKED ISSUE #566: [CRITICAL] backslash behind tickets URL crashes site
The link https://areafiftylan.nl/tickets/ is in the email. It will crash the site. https://areafiftylan.nl/tickets works though.

Happens in both Chrome and Firefox.

image

Fastjur:
Fixes #566

### DIFF
--- a/docker/nginx_server.conf
+++ b/docker/nginx_server.conf
@@ -4,6 +4,8 @@ server {
     root         /srv/www/;
     index        index.html;
 
+    rewrite ^/(.*)/$ /$1 permanent;
+
     location / {
         try_files $uri /index.html;
     }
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Probably the following need to be updated as well:

1. data-path of all iron-lazy-pages sub-pages
2. The url of `index-imports` which is lazy-loaded

TimvdLippe:
And we should fix up our nginx config to not return index.html when a `*.{js,html,css,png}` request is not found. That should throw a 404 instead. I did fix that in Apache config back then, but needs to be ported to nginx now.

PR #567, opened 2018-01-11
New banner design

martijnjanssen:
The slider was adding nothing to the webpage other than slower load times, we agreed to remove it.

The tabs at the bottom of the banner are clickable when there is an underline under the text, allowing the users to navigate quickly after reading the message.

image


index.html
```diff
@@ -119,7 +119,7 @@
       border-bottom: 2px solid var(--secondary-color);
     }
 
-    img {
+    .headerLogo {
```
martijnjanssen: This css property gave EVERY img item in the webpage a margin on the left, reaping havoc all over. I removed it 😛

Fastjur: Woopteedoo. Good catch

martijnjanssen:
Huh, I thought I made it for smaller screens, at least 360px. I'll check.

Fastjur:
I think the text I changed breaks it. The wrapping makes the text go above and below the yellow block

martijnjanssen:
This is the updated banner right now, on a wide and a narrow screen:

screen shot 2018-01-23 at 14 04 37
screen shot 2018-01-23 at 14 05 05

PR #571, opened 2018-01-15
Lastpass workaround, fix loading the page

martijnjanssen:
The webpage now loads with lastpass enabled, using [this](https://gist.github.com/usergenic/9baf88a24dcbc77bc808339755a7d436) workaround. I couldn't use the html import for some reason, so I added it before the js import.

### DIFF
--- a/ce-fix.html
+++ b/ce-fix.html
@@ -1,0 +1,3 @@
+<script>
+  document.createElement = Document.prototype.createElement;
+</script>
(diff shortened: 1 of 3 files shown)


martijnjanssen:
I don't know why, but it just doesn't work when put in the `script` in the index.html at all. I put it at line 327, at the first line of the lambda, directly after the `'use strict;` statement and as an html import in the header. This is the only way it works, the other ways just don't work 😞 

So for the polyfill, it would be preferrable to fix it there? You know that by doing that it will remove the ability for the lastpass team to reproduce the issue and fix it there?

elarb:
Did you make sure its first html import encountered when trying to import it in the header?

PR #573, opened 2018-01-16
Add service worker

### LINKED ISSUE #474: Show toast when site has been updated
When a new version of the website is published and a new service-worker is registered, we should notify the user that is the case. For example https://github.com/webcomponents/webcomponents.org/blob/f7027235caece5b912aea04a6ea34ef05449ccc0/client/src/lazy-block.html#L101-L108 fires an event when the SW is registered, it shows a nice toast.

elarb:
- Adds a Service Worker to provide better user experience in flaky network situations and offline
- Added another `paper-toast`  that should open when we want the user to reload the page (e.g. sw update, client error, ...)

Fixes #474


index.html
```diff
@@ -150,7 +150,11 @@
 
 <body>
   <lancie-auth-storage></lancie-auth-storage>
-  <paper-toast></paper-toast>
+  <paper-toast id="msg-toast"></paper-toast>
+
+  <paper-toast id="reload-toast">
+    <paper-button onclick="window.location.reload()">Reload</paper-button>
```
TimvdLippe: Let's use a normal `<button>` for this. Since we reload the page right away, we don't need any fancities. (Actually, a `<a>` is nicer here in the toast I think)

scripts/sw-message.js
```diff
@@ -0,0 +1,5 @@
+self.addEventListener('message', function(event) {
```
TimvdLippe: Why is this file here?

scripts/sw-register.js
```diff
@@ -0,0 +1,39 @@
+(() => {
+  if ('serviceWorker' in navigator && window.location.hostname !== 'localhost') {
```
TimvdLippe: We don't need to filter out localhost as we have a no-op SW anyways.

scripts/sw-message.js
elarb: It gets imported in the generated `service-worker.js`; skips waiting when the message is being sent to the Service Worker

scripts/sw-register.js
```diff
@@ -0,0 +1,39 @@
...
+
+        reg.addEventListener('updatefound', function () {
+          trackInstalling(reg.installing);
+        });
+      }
+    )
+  }
+
+  let updateReady = (worker) => {
+    this.fire('update-toast');
```
TimvdLippe: I think this should be `this.dispatchEvent(new CustomEvent(...))`, as `fire` is a Polymer method

index.html
```diff
@@ -150,7 +150,11 @@
 
 <body>
   <lancie-auth-storage></lancie-auth-storage>
-  <paper-toast></paper-toast>
+  <paper-toast id="msg-toast"></paper-toast>
+
+  <paper-toast id="reload-toast">
+    <a href=".">Reload</a>
```
martijnjanssen: This link shows above the header when the page is loading after a refresh.

elarb:
Still WIP! Needs some further testing

elarb:
Works as expected now:

sw

Could someone else confirm this too?

TimvdLippe:
@elarb Hm, I am unable to receive the popup. The service-worker gets installed, but the popup does not show up after reloading. Moreover, when I go offline none of the files are retrieved, which makes me wonder whether the scope of the service-worker is actually correct. Since the file resides in modern/ or fallback/, maybe the service-worker is not correctly triggered?

In total, I have seen the popup once, but not consistently. I am using `prpl-server` from `build/` to serve. I have also added the scope to `navigator.serviceWorker.register('service-worker.js', {scope: '/'}).then((reg) => {` per https://github.com/Polymer/prpl-server-node#scope-header, but that did not fully fix it either.

Fastjur:
Is the idea of the reload popup for when files have changed? Thus for the developer?

TimvdLippe:
@Fastjur for users to be notified when the site has updated content.

Fastjur:
How does it get triggered then? Only due to updates of the service-worker.js?

How does this work live?

TimvdLippe:
Whenever content of the website changes, the hashes as used in the SW change. This change in the binary of a SW triggers the update.

E.g. developer changes file, build process generates new SW with new hashes, this triggers SW update, users receive popup

Fastjur:
I see.

I don't think we have that many users live on our site though whenever we update? It's a very cool feature nonetheless. Especially now that we can give users a usable reload toast.

Fastjur:
I cannot get a popup working though. Whenever I change the service worker I do not get a toast.

TimvdLippe:
> I don't think we have that many users live on our site though whenever we update?

This is not about being live at the same time. This update triggers as soon as you visit the website when 1. you visited the website before the update 2. the update was released and 3. you have not visited the website after the update. Then the website detects an update was released compared to the last time you visited and shows the popup

> I cannot get a popup working though. Whenever I change the service worker I do not get a toast.

Yes that is correct, see my comment at https://github.com/AreaFiftyLAN/lancie-frontend/pull/573#issuecomment-361061287

Fastjur:
I see. Why is a reload sensible in that case?

TimvdLippe:
Because the old SW still controls the website. You need to trigger a reload to essentially say: hey, new SW you take control now.

elarb:
It took quite some time debugging, but I think I got it to work (with prpl-server too). Could someone confirm this once again?

PR #589, opened 2018-01-28
Move dockerbuild into DockerFile for faster container builds

TimvdLippe:
With this change, running `docker build` will now cleverly use the cache of installed global npm modules. This should make building a lot faster when you are just editing source files.

Also update the instructions in the README and remove some unused files that should not end up in the container.


praseodym:
A general recommendation would be to run the `prpl-server` process as a user other than `root` (will require the listen port to be changed to a non-privileged port like 8080).

Besides, I'm not quite sure why the `COPY . .` is needed:
https://github.com/AreaFiftyLAN/lancie-frontend/blob/78b1449750137a078fe48d63d4bd7859c906d8bb/Dockerfile#L9

TimvdLippe:
@skamoen @martijnjanssen Yay/nay? I think it follows the same pattern as described in https://blog.hasura.io/an-exhaustive-guide-to-writing-dockerfiles-for-node-js-web-apps-bbee6bd2f3c4

praseodym:
I think my comments are still valid.

elarb:
@praseodym Thanks, I applied your suggestions, could you review it please?

Dockerfile
```diff
@@ -1,11 +1,28 @@
...
+COPY index.html ce-fix.html favicon.ico gulpfile.js manifest.json polymer.json robots.txt .jshintrc ./
+RUN yarn run build
 
 FROM node:8-alpine
 RUN npm install -g prpl-server
-COPY --from=builder /src/build/ .
-COPY . .
-EXPOSE 80
-CMD [ "prpl-server", "--config" , "polymer.json" ,"--host" ,"0.0.0.0", "--port" ,"80" ]
+COPY --from=builder /app/build/ .
```
praseodym: `.` is actually the root (`/`) of the container here, which is a little messy. I'd suggest adding a `WORKDIR /` before this.

Dockerfile
```diff
@@ -1,9 +1,16 @@
...
-RUN yarn
+
+RUN useradd -r -m lancie
+RUN mkdir /app
+RUN chown lancie /app
+
 RUN npm install -g bower
 RUN npm install -g polymer-cli --unsafe-perm
+
+USER lancie
```
praseodym: This (as well as the `useradd`) needs to go in the final container, not the builder container.

Dockerfile
```diff
@@ -1,9 +1,16 @@
 FROM node:8 AS builder
-WORKDIR /src
-COPY package.json yarn.lock ./
-RUN yarn
+
+RUN useradd -r -m lancie
```
praseodym: I'd suggest also adding a group:
`RUN groupadd -r lancie --gid=999 && useradd --no-log-init -r -g spring-boot --uid=999 lancie`

Dockerfile
```diff
@@ -1,9 +1,16 @@
...
+RUN mkdir /app
+RUN chown lancie /app
```
praseodym: `chown` is not needed (and discouraged) for the final container unless temporary files are created.

praseodym:
The comments marked as 'outdated' are actually still relevant (I just accidentally posted them on a previous commit)

elarb:
@praseodym Pinging you once more, could you confirm its done properly now?

PR #590, opened 2018-01-28
workaround for `paper-input` with type"date" not being editable on iOS

### LINKED ISSUE #585: Birthday not editable when empty on mobile
At least on safari on iOS

elarb:
Fixes #585


elarb:
Tested with BrowserStack on iOS and can confirm that it fixed the issue

PR #593, opened 2018-01-29
Feature dynamic seat groups

Fastjur:
* Generates the seatgroups dynamically instead of having to set them manually in a string
* Fixes the taken -> locked bug (red colours now show up)


src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -160,13 +161,14 @@ <h2>Seat reservation</h2>
...
       onGetAllSeats: function(e, request) {
         if (request.succeeded) {
-          var seats = [];
-          var groups = 'ABCDEFGHIJ';
-          for (var i = 0; i < groups.length; i++) {
-            seats[groups.charAt(i)] = request.response.seatmap[groups.charAt(i)].sort(function(a, b) {
+          let seats = [],
+              groups = Object.keys(request.response.seatmap);
+          this.set('seatGroups', groups);
+          groups.map(group => {
```
TimvdLippe: Should be `forEach`

Fastjur: ...

Sometimes I question whether I am in the right discipline...

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -3,6 +3,8 @@
 Copyright (c) 2015 The LANcie commission of W.I.S.V `Christiaan Huygens`. All rights reserved.
 This code may only be used under the BSD style license found at https://github.com/WISVCH/LANcie/blob/master/LICENSE
 -->
+<link rel="import" href="../../bower_components/iron-icons/iron-icons.html">
```
elarb: Lets add the icon needed to the custom icon set (`lancie-icons.html`) instead? You could simply copy and pase

Fastjur: Yes good point :+1:

src/lancie-seatmap/lancie-seat.html
```diff
@@ -86,40 +94,46 @@
         }
       },
       showSeatDialog: function() {
-        if (this.seat.taken) {
+        if (this.seat.locked || this.seat.ticket && this.seat.ticket.owner.reference !== this.user.reference) {
```
elarb: Extra parentheses would be a good idea, e.g. :
(true || false && false) -> true
((true || false) && false) -> false

Fastjur: The logic actually is:

`this.seat.locked`

OR

`this.seat.ticket && this.seat.ticket.owner.reference !== this.user.reference`

There is no need for the parentheses here

Fastjur: The `this.seat.ticket` is there to prevent checking the owner reference on an `undefined` ticket, causing the seat to not be rendered at all

Fastjur:
This is still a WIP. Current state:

image

I could refactor to ES6 if we want to.

I'll give you a ping once I deem this is ready to review again, sounds good?

elarb:
@Fastjur Sure!

Fastjur:
I need your opinion on this guys.  
Because the seats etc are now a static size, they cannot easily adjust to the viewport.

This is no problem, however the seatmap then isn't centered and not everything is filled to the right on bigger screens.

I came up with this:

image

This however does not always show clearly that there is more content to the right.

Any ideas on fixing this?

elarb:
@Fastjur 
Why does the size have to be static? 

Also, what should happen when clicking on a non-free spot? I am getting this on the first click:
screenshot from 2018-01-31 13-32-58

elarb:
I made some small changes. Can you confirm it still behaves as it should? @Fastjur (I couldn't spot anything different)

TimvdLippe:
Reserving a seat appears to be broken. I get a 400 when I click on a seat and then a user. Also the popup immediately hides, while the warning should still be there. Lastly, the warning should be hidden once the popup is closed, to prevent the state from "flowing over" to a new click.

TimvdLippe:
image
If there are special seats in a seat group, the ordering of the numbers goes wrong. It should always start at 1.

TimvdLippe:
I think the 400 happened because the user was already assigned to a seat. In this case, the user should swap to the new seat. Is this an API or frontend issue?

skamoen:
Reserving seats seems to work fine. However I really don't like the scrollbar. It's not obvious there are more seats and people will get confused by this. This needs a better solution.

Fastjur:
@skamoen yeah I had a discussion with some people. Perhaps something with 2 arrow like buttons on the side?

Fastjur:
@elarb To answer your question why I wanted it static (at least this year) is due to the fact that rows will not have the same amount of seats this year, causing some seats to appear 'smaller'. The easiest fix was to make the seats static. We might have to come up with a better solution to make it both fill the viewport, and still have the seats appear the same size. But that is something for after this years LAN

src/lancie-seatmap/lancie-seat.html
```diff
@@ -83,43 +92,56 @@
...
         }
-        if (!this.fixed && user.reference === seat.ticket.owner.reference) {
-          return 'green';
-        }
-        var teamMembers = teamTickets.map(function(ticket) {
-          return ticket.owner.reference;
-        });
-        if (!this.fixed && teamMembers.includes(seat.ticket.owner.reference)) {
-          return 'yellow';
+        if (seat.ticket) {
```
martijnjanssen: Why not do the `!seat.ticket` case first? This will give easier readable code with less if-case nesting 👍

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -78,118 +117,145 @@
...
-        </div>
+    <div class="seatmapwrapper">
+      <div class="seatmap">
+         <div class="seats">
+            <template is="dom-repeat" items='[[seatGroups]]'>
+               <div class="seatmap-row">
+                  <div class="grouplabel">[[item]]</div>
+                  <div class="seats-group">
+                     <template is="dom-repeat" items="[[_getSeatGroup(seats, item)]]">
+                        <lancie-seat tickets="{{tickets}}" filter="[[filterUser]]"  seat="{{item}}" user="{{user}}" fixed="{{fixed}}"></lancie-seat>
```
martijnjanssen: Does this all need double binding?

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -200,18 +266,22 @@ <h2>Seat reservation</h2>
             this.$.error.setError('Something went wrong, please refresh and try again.');
           }
         }
-      },
-      open: function(event) {
+      }
+
+      open(event) {
```
martijnjanssen: Let's change this function to `openReservationDialog`

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -78,118 +117,145 @@
...
+        for (let group of seats) {
+          group.sort((a, b) => a.seatNumber - b.seatNumber);
+        }
+
+        this.seats = seats;
+      }
+
+      onGetTeamTickets(e, request) {
+        if (!request.succeeded) {
+          this.$.toast.show({
```
martijnjanssen: We can change this to a `refreshToast` event!

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -78,118 +117,145 @@
...
-          Select the user to reserve this seat for:
-          <paper-listbox id="listbox">
-            <template is="dom-repeat" items="[[tickets]]">
-              <paper-item data-ticket$="[[item.id]]">[[item.owner.profile.displayName]]</paper-item>
+  <lancie-dialog id="seatDialog">
+    <h2>Seat reservation</h2>
+    <div class="dialog-content">
+      <lancie-error id="error"></lancie-error>
+      <div hidden$="[[!_checkExists(tickets)]]">
+        Select the user to reserve this seat for:
```
martijnjanssen: Wrapping the text in a `<p>` element improves styling.

src/lancie-seatmap/lancie-seatmap.html
```diff
@@ -78,118 +117,145 @@
...
-        <div hidden$="[[_checkExists(tickets)]]">
-          You need at least one ticket to reserve a seat!
-        </div>
+        </paper-listbox>
       </div>
-      <div class="dialog-actions">
-        <paper-button dialog-dismiss>Close</paper-button>
-        <paper-button on-tap="reserveSeat" hidden$="[[!_checkExists(tickets)]]">Reserve seat</paper-button>
+      <div hidden$="[[_checkExists(tickets)]]">
+        You need at least one ticket to reserve a seat!
```
martijnjanssen: This should also be better in a `<p>` element.

PR #596, opened 2018-02-03
Remove early birds mention from homepage

TimvdLippe:


elarb:
Maybe change it to something like "Selling out fast"?

PR #606, opened 2018-02-13
Add link back to the homepage to the construction page

TimvdLippe:
Better for navigation of the user.


src/lancie-construction.html
```diff
@@ -21,6 +21,7 @@
     <lancie-card heading="Coming soon!">
       <div class="card-content">
         This page in under construction and will be made available soon. Please come back later!
```
martijnjanssen: "This page in under", shudders...

PR #614, opened 2018-02-15
Update tournaments

Fastjur:
screenshot from 2018-02-16 12-23-01

## League of Legends 1v1 (unofficial)
screenshot from 2018-02-16 12-23-21
screenshot from 2018-02-16 12-23-32

## Dota 2 5v5 (unofficial)
screenshot from 2018-02-16 12-23-49
screenshot from 2018-02-16 12-24-00

## Trackmania (Official, added TransIP sponsor, no prizes yet)
screenshot from 2018-02-16 12-24-22
The second and third prize are the same, it is not a typo. Not certain how I think about the way that looks though.


Fastjur:
Problem is that we do not know exactly how they will be split, as we have quite a lot of TM keys and goodies. They might be used in other tournaments as small prizes. So mentioning them here doesn't seem smart IMO.

PR #630, opened 2018-03-21
Initial my-area layout update

elarb:
Initial my-area layout update:
- Used custom icon set project-wide
- Added info boxes to lancie-area:
info-boxes
- Removed the separate element for discord


src/index-imports.html
```diff
@@ -11,7 +11,6 @@
 <link rel="import" href="../bower_components/app-layout/app-drawer/app-drawer.html">
 <link rel="import" href="../bower_components/paper-listbox/paper-listbox.html">
 <!-- TODO: Generate custom icon set -->
```
TimvdLippe: We can remove this comment now

src/lancie-my-area/lancie-info-box.html
```diff
@@ -0,0 +1,73 @@
...
+      :host {
+        display: flex;
+        flex-direction: row;
+        justify-content: space-between;
+        margin: 8px;
+        padding: 20px;
+        background: var(--base-bg-color);
+        box-shadow: 0 2px 5px 0 rgba(0, 0, 0, .26);
+        --icon-fill-color: var(--paper-grey-700);
+        @apply --layout-flex;
```
TimvdLippe: Do not use apply but regular css

src/lancie-my-area/lancie-my-area.html
```diff
@@ -15,43 +17,124 @@
...
+      @apply --layout-horizontal;
+      @apply --layout-wrap;
+      @apply --layout-start;
+    }
+
+    .cards-container > div {
+      @apply --layout-vertical;
+      @apply --layout-wrap;
+      @apply --layout-flex;
       margin: 8px;
```
TimvdLippe: Do not use apply but regular css

src/lancie-my-area/lancie-my-area.html
```diff
@@ -15,43 +17,124 @@
...
+
+    .discord {
+      --icon-fill-color: var(--paper-grey-800);
+    }
+
+    .discord:hover {
+      cursor: pointer;
+      -webkit-transform: translateY(-2px);
+      transform: translateY(-2px);
+      box-shadow: 0 7px 14px rgba(50, 50, 93, .1), 0 3px 6px rgba(0, 0, 0, .08)
```
TimvdLippe: Lots of magic values here. Will have to look what the visuals are

src/lancie-my-area/my-area-orders.html
```diff
@@ -130,12 +136,9 @@
         })(this));
         this.tickets = this.tempTickets;
       } else {
-        this.$.ticketTransferTokensRetrievalError.show();
+        this.fire('toast', {text: 'Unable to check for open ticket transfers!', duration: 3000});
```
TimvdLippe: Should be dispatchevent instead of fire

src/lancie-my-area/lancie-my-area.html
```diff
@@ -15,43 +17,124 @@
...
+    </lancie-ajax>
+
     <lancie-section type="primary full" header="My Area - {{user.profile.displayName}}">
-      <div>
-        <div class="layout horizontal start wrap">
-          <my-area-profile user="{{user}}" class="flex-2 card"></my-area-profile>
-          <my-area-teams user="{{user}}" class="flex-2 card"></my-area-teams>
-          <my-area-discord class="flex-1 card"></my-area-discord>
+      <div class="info-boxes">
+        <lancie-info-box class="seat" heading="Seat" icon="lancie:event-seat" info$="{{_computeSeatInfo(seats)}}"></lancie-info-box>
```
Fastjur: Why not make this one link to `/seatmap`? So the buttons have some navigational purpose too where possible

src/lancie-my-area/lancie-my-area.html
Fastjur: We could also use the actual style classes in the `style=` tag right?

src/lancie-my-area/lancie-my-area.html
Fastjur: Magic values for a box shadow are hard to circumvent tho?

src/lancie-my-area/my-area-orders.html
elarb: This element is still using the `Polymer` factory method

src/lancie-my-area/lancie-info-box.html
```diff
@@ -0,0 +1,73 @@
...
+
+      .info {
+        font-size: 1.8em;
+        font-weight: 500;
+        margin-top: 5px;
+      }
+
+    </style>
+
+    <iron-icon icon="{{icon}}"></iron-icon>
```
martijnjanssen: Why double bindings?

src/lancie-my-area/lancie-my-area.html
martijnjanssen: Yeah, but these values are different from every other place where we use a box shadow. Is there a specific reason for this.

src/lancie-my-area/my-area-orders.html
martijnjanssen: We don't have to specify a duration here, the default is already this value.

src/lancie-my-area/lancie-my-area.html
```diff
@@ -15,43 +17,126 @@
...
+      _computeSeatInfo(seats) {
+        for (let seat of seats) {
+          if (seat.ticket.owner.reference === this.user.reference) {
+            return `${seat.seatGroup}${seat.seatNumber}`;
+          }
+        }
+      }
+
+      _computeTicketInfo(tickets) {
+        return tickets ? (tickets[0].valid ? 'Purchased' : 'Unknown') : 'None';
```
martijnjanssen: I think we have to add an if for the case that the tickets array is empty.

Fastjur: `tickets && tickets.length > 0` should work right?

As when tickets does not exist the second part does not get evaluated

TimvdLippe: `return (tickets && tickets.length) ? (tickets[0].valid ? 'Purchased' : 'Unknown') : 'None';`

src/lancie-my-area/lancie-my-area.html
martijnjanssen: I think that linking it to seatmap without a button is a bad idea. It's not actually a button, it's an info card.

src/lancie-my-area/lancie-my-area.html
elarb: infobox

src/lancie-my-area/lancie-my-area.html
```diff
@@ -1,57 +1,161 @@
...
+      absoluteurl="https://discordapp.com/api/servers/146690066541051904/widget.json"
+      last-response="{{discordWidget}}">
+    </lancie-ajax>
+
+    <div class="container">
+      <h1>My Area - [[user.profile.displayName]]</h1>
+      <div class="info-boxes">
+        <lancie-info-box class="seat" heading="Seat" icon="lancie:event-seat" info$="{{_computeSeatInfo(seats)}}"></lancie-info-box>
+        <lancie-info-box class="ticket" heading="Ticket" icon="lancie:local-activity" info$="{{_computeTicketInfo(tickets)}}"></lancie-info-box>
+        <lancie-info-box class="team" heading="Team" icon="lancie:group" info$="{{_computeTeamInfo(teams)}}"></lancie-info-box>
```
TimvdLippe: These `info` should not be attributes, e.g. remove the `$`. Also should they be two-way-binding?

elarb: The way I initially created these `info-boxes` was slightly different, will fix this

src/lancie-my-area/lancie-my-area.html
```diff
@@ -68,7 +76,17 @@
       cursor: pointer;
       -webkit-transform: translateY(-2px);
       transform: translateY(-2px);
-      box-shadow: 0 7px 14px rgba(50, 50, 93, .1), 0 3px 6px rgba(0, 0, 0, .08)
+      box-shadow: 0 7px 14px rgba(50, 50, 93, .1), 0 3px 6px rgba(0, 0, 0, .08);
```
martijnjanssen: I'd replace this one with: `box-shadow: 0 14px 21px rgba(50, 50, 93, .1), 0 6px 9px rgba(0, 0, 0, .08);` such that it appears that it raises.

src/lancie-my-area/lancie-my-area.html
```diff
@@ -68,7 +76,17 @@
       cursor: pointer;
       -webkit-transform: translateY(-2px);
       transform: translateY(-2px);
```
martijnjanssen: The transform of `2px` is a bit strange, it looks way better with the new shadow and without the tranform.

src/lancie-my-area/lancie-my-area.html
elarb: Could you make a commit with this change, please?

PR #631, opened 2018-03-27
Fix image optimizer for windows file paths

martijnjanssen:


gulpfile.js
```diff
@@ -45,7 +45,7 @@ const optimizeImages = () => new Promise((resolve, reject) => {
       if (path.extname(file) === '.svg') {
         fs.copySync(file, path.resolve(optimizedImagesRoot, relativeFile));
       } else {
-        const imageCategory = path.relative(root, file).split('/')[0];
+        const imageCategory = path.relative(root, file).split(path.sep)[0];
```
TimvdLippe: Please do the same for line 43

elarb:
Aha you found the culprit, nice!

PR #632, opened 2018-03-27
Refactor profile edit form

### LINKED ISSUE #582: Profile form: gender, birthday and phone number fields too large

elarb:
- Use ES6 syntax
- Improve styling

Closes #582


skamoen:
There's no real reason why we need to know the Gender, so let's remove the field entirely. This needs a change in the API as well, so send "Other" by default until we fix that.

elarb:
@skamoen I think its better to do that in another PR and immediately change it in the API as well

src/lancie-my-area/lancie-profile-edit.html
```diff
@@ -0,0 +1,171 @@
...
+      }
+
+      .dropdown-content {
+        min-width: 120px;
+      }
+
+      /*
+       Workaround for a bug that causes paper-input
+       with type="date" not being editable when empty
+       (PolymerElements/paper-input/issues/352).
```
martijnjanssen: I think this is no longer needed when updating `paper-input` to `2.2.0` according to [this](https://github.com/PolymerElements/paper-input/commit/cc91da23163e682b67f32007ba27a007b7b7cd3d).

elarb: We updated to `2.2.0` before this issue appeared

martijnjanssen: We're still on [2.0.0](https://github.com/AreaFiftyLAN/lancie-frontend/blob/master/bower.json#L29) :wink:

skamoen:
Agreed on the gender removal in a separate PR. 

Travis is unhappy with a very lengthy error log. Can you look into this?

martijnjanssen:
Looks good, in my commit I fixed:
- The next button when no profile changes are made in the ticket page
- Close the profile edit form when the user clicks on the check in my-area

PR #635, opened 2018-03-28
Collapse terms and conditions in ticket checkout

### LINKED ISSUE #616: Having to scroll past the entire T&C is annoying
We should think about placing the T&C inside a div and letting that scroll, while keeping the buttons in the screen. The current layout negatively impacts mobile users, the scroll distance is gigantic.

martijnjanssen:
Fixes #616


skamoen:
This is not some ridiculous EULA. People should actually read this. Not sure if this is desired.

martijnjanssen:
Okay let's close this, since this is not exactly the thing we want.

PR #640, opened 2018-04-28
Add `lancie-icons` to fragments

elarb:
Icons didn't load after building, this should fix it.


TimvdLippe:
Hm, I don't think this should be its own fragment. Rather it should be an import of `index.html`, or we have to have our imports correctly such that it ends up in the shared bundle.

PR #644, opened 2018-09-29
Extract password edit from profile

martijnjanssen:
This is a first quick change, I still have some ideas for my-area, so the exact styling is not that important here I think.

### DIFF
--- a/src/lancie-my-area/my-area-password.html
+++ b/src/lancie-my-area/my-area-password.html
@@ -11,117 +11,135 @@
 <link rel="import" href="../lancie-icons.html">
 
 <dom-module id="my-area-password">
-    <template>
-        <style include="iron-flex iron-flex-alignment"></style>
-        <style>
-            :host {
-      display: block;
-    }
+<template>
+<style>
+  :host {
+    display: block;
+  }
 
-    [hidden] {
-      display: none !important;
-    }
+  [hidden] {
+    display: none !important;
+  }
 
-    paper-tooltip {
-      --paper-tooltip-background: var(--primary-color);
-      --paper-tooltip-text-color: var(--secondary-color);
-    }
+  paper-card {
+    display: block;
+  }
 
-    paper-icon-button[icon="lancie:clear"] {
-      --iron-icon-fill-color: #F44336;
-    }
+  .horizontal {
+    display: flex;
+    align-items: center;
+  }
 
-    paper-icon-button[icon="lancie:check"] {
-      --iron-icon-fill-color: #4CAF50;
-    }
+  paper-tooltip {
+    --paper-tooltip-background: var(--primary-color);
+    --paper-tooltip-text-color: var(--secondary-color);
+  }
 
-    iron-icon {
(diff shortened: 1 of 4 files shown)


src/lancie-my-area/my-area-password.html
```diff
@@ -0,0 +1,127 @@
+<link rel="import" href="../../bower_components/polymer/polymer.html">
```
elarb: Maybe its better if we use ES6 class-based syntax for this new element?

src/lancie-my-area/my-area-password.html
```diff
@@ -0,0 +1,127 @@
...
+<link rel="import" href="../../bower_components/paper-tooltip/paper-tooltip.html">
+<link rel="import" href="../../bower_components/paper-toast/paper-toast.html">
+
+<link rel="import" href="../../bower_components/lancie-form/lancie-form.html">
+<link rel="import" href="../../bower_components/lancie-login-form/lancie-passwords.html">
+<link rel="import" href="../lancie-icons.html">
+
+<dom-module id="my-area-password">
+    <template>
+        <style include="iron-flex iron-flex-alignment"></style>
```
elarb: Let's use flexbox directly to style the content?

src/lancie-my-area/my-area-password.html
martijnjanssen: Is it the ES6 class-based syntax now?

elarb:
Just tested it (both with a correct and incorrect password), seems to be broken. Doesn't give any feedback about changes and when inspecting the network the response is a 403 (Access Denied).

martijnjanssen:
Yeah, the 403's were because it did work the first time, but upon the second time trying, the password was already changed. The things that are broken are the events.

martijnjanssen:
Could someone test this? I’m pretty sure that everything is fixed here.

timovanasten (changes requested):
Works when correct password is filled in and new password typed correctly twice. 

When its not supposed to work because the password is not correct or the new password does not meet the password requirements (although that is currently not running, it's already in master on the API side), it is not clear to the user why because it just states "Password update failed".

I don't think we should merge until some sort of explanation of the password requirements is in here and the reason of update failure is made clear to the user. I can imagine that that would require changes to the API.

ArdyZ:
@timovanasten and @martijnjanssen can you have a look at this pull request?

timovanasten:
@ArdyZ Boy, that is a while ago haha. I believe the code in this PR adding the functionality to change passwords was functioning (at the time at least), but I requested some changes from a user experience point of view. The API enforces some constraints on the password, but these are not communicated to the user. A simple piece of text with the password requirements should suffice. For example, something like "Password should be <x> characters long and ....". 

I don't have the environment set up to work on this project atm, so if you could verify that this still works and add the text with the password requirements to the element then this could be merged if you ask me.

Fastjur:
I'm just commenting here so I can be part of the history before this thread is closed :P

PR #645, opened 2018-09-30
Simple my-area layout idea

martijnjanssen:
Is this maybe a different layout that would be better? Maybe this in combination with a full drawer layout, allowing us to make a google inbox like experience for mobile and desktop. We can divide my-area in different sub-pages in my-area, using the same menu structure in the drawer:

- Home
- Tickets
- My-Area
  - Profile
  - Password
  - Teams
  - Tickets
  - Seat
- Contact

To expand this design for my-area further, each team will have it's own paper-card in the teams page. The same for the tickets, one paper-card for each ticket.

### DIFF
--- a/src/lancie-my-area/lancie-my-area.html
+++ b/src/lancie-my-area/lancie-my-area.html
@@ -21,29 +21,19 @@
       line-height: 48px;
       margin: 0 0 16px 8px;
     }
+
     .container {
       padding: 32px;
     }
 
-    .cards-container {
-      display: flex;
-      flex-direction: row;
-      flex-wrap: wrap;
-      justify-content: flex-start;
+    .action-cards {
+      max-width: 800px;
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Could you add some screenshots?

martijnjanssen:
screen shot 2018-10-01 at 21 00 37
screen shot 2018-10-01 at 21 00 10

ArdyZ:
@martijnjanssen what should happen to this pull request to finish it?

martijnjanssen:
I think the main idea here was to change the tile-layout for My-Area into sub-pages. This pr changes to a single column I believe. The next steps (if this feature is event desired) would be to use something like `paper-tabs` and then have the page-routing change accordingly. However, it's just a feature, not something that is really needed :)

PR #646, opened 2018-10-07
Add a way to log out

elarb:
This PR introduces the long awaited feature enabling users to logout


index.html
```diff
@@ -317,12 +326,26 @@
           el.user = user;
         }
       }
+      function showLogoutButtons() {
+        for (const node of logoutButtons) {
+          node.style.display = 'block';
```
TimvdLippe: Let's add a class rather than inline style. E.g. `node.classList.add('signed')` or something

elarb:
Test

src/lancie-login/lancie-logout-button.html
```diff
@@ -0,0 +1,58 @@
...
+        return 'lancie-logout-button';
+      }
+
+      static get properties() {
+        return {
+          labeled: {
+            type: Boolean,
+            value: false,
+          }
+        }
```
martijnjanssen: Missing semicolumn here: causing the build to fail
```suggestion
        };
```

src/lancie-login/lancie-logout-button.html
```diff
@@ -0,0 +1,58 @@
...
+
+    <lancie-ajax id="ajaxLogout" refurl="logout" method="POST" on-lancie-ajax="onLogoutResponse"></lancie-ajax>
+    <div on-tap="tryLogout">
+      <paper-icon-button icon="lancie:exit-to-app"></paper-icon-button>
+      <span hidden="[[!labeled]]">Sign out</span>
+    </div>
+
+  </template>
+  <script>
+    class LancieLogoutButton extends Polymer.Element {
```
martijnjanssen: Also need to encapsulate the js of the element in a function declaration.

elarb: I don't think that's necessary

PR #651, opened 2018-12-20
Feature content update

timhuisman1:
Website content update to be used for next LAN :)

### DIFF
--- a/data/official-tournaments.json
+++ b/data/official-tournaments.json
@@ -11,7 +11,7 @@
       "website": "https://www.rocketleague.com"
     },
     "prizes": [
-      "Steam Controller (3x)"
+      "To be announced"
     ]
   },
   {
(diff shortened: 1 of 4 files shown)


data/official-tournaments.json
```diff
@@ -11,7 +11,7 @@
       "website": "https://www.rocketleague.com"
     },
     "prizes": [
-      "Steam Controller (3x)"
+      "To be announced"
```
martijnjanssen: Sure, but why not remove the tournaments all together? There is now a way to set and get them from the api!

timhuisman1: I thought that was not working correctly yet, but anyway I mainly just want the website to be updated so that we could start promoting. I will fix it in the future, probably when we know more about sponsors and prizes.

martijnjanssen:
I'll merge this! One small comment, I'd recommend bumping the official and unofficial tournaments json in the `lancie-home-page`. In the past we had some issues with this file being cached on the users' side and the user not seeing the updated content. With the version bump, the file is seen as updated and the user will see the updated one from the server.

package.json
```diff
@@ -22,5 +22,8 @@
   },
   "engines": {
     "node": ">=5.0.0"
+  },
+  "dependencies": {
+    "polymer-cli": "^1.9.1"
```
martijnjanssen: `polymer-cli` is not supposed to be here. It should be installed globally.

PR #653, opened 2018-12-24
Banner image and icon update

timovanasten:
Increased the resolution of the banner image and removed the circles that were never intended to be there (in the design they were placeholders for icons I was told by @svenpopping). I've replaced them with actual icons.

### Before 
before
### After
after


TimvdLippe:
The first two commits should not be there.

src/lancie-home-page/lancie-info-blocks.html
```diff
@@ -11,21 +12,10 @@
...
   }
 
-  .container > div::before {
-    min-width: 36px;
-    min-height: 36px;
-    border-radius: 50%;
-    margin: auto 16px auto 0;
-    border: 4px solid var(--secondary-color);
-    background: transparent;
-    content: "";
```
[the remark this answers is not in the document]
timovanasten: This used to be for the circles that I changed into icons

elarb:
Looks good! What do you think about using the ticket icon (`local-activity`) instead of the euro icon?

screenshot from 2018-12-25 05-42-14

timovanasten:
I like it, will do

src/lancie-home-page/lancie-info-blocks.html
```diff
@@ -11,21 +13,10 @@
   .container {
     display: flex;
     flex-wrap: wrap;
-    max-width: 1032px;
```
martijnjanssen: Is it not strange that the blocks are now wider than the site content.

timovanasten: I don't think these blocks need to be the same width as the other site content. At least I didn't like the restriction of 1032px as shown below. It leaves too little spacing between the blocks in my opinion, which is unneeded if you have room on the sides.

schermafbeelding 2018-12-25 om 13 26 58

I do agree that some sort of width restriction is nice. I like the spacing on my monitor which is 1440px. How about that?

PR #662, opened 2019-01-25
Increases the order limit from 5 to 15

timovanasten:


src/lancie-ticket-page/lancie-ticket-item.html
```diff
@@ -105,6 +105,8 @@
   'use strict';
 
   (function() {
+    const orderLimit = 15;
```
martijnjanssen: I'd much prefer if you declare the orderLimit in the properties. You can do that by adding:

```
orderLimit: {
  type: Number,
  value: 15,
},
```

src/lancie-ticket-page/lancie-ticket-item.html
```diff
@@ -141,8 +143,8 @@
       addToBasket: function() {
         if (!this.basket.tickets) {
           this.$.addTicket.refurl = 'orders';
-        } else if (this.basket.tickets.length >= 5) {
-          this.fire('toast', {text: 'Maximum of 5 tickets in order.'});
+        } else if (this.basket.tickets.length >= orderLimit) {
```
martijnjanssen: And this would then become: `this.orderLimit`

src/lancie-ticket-page/lancie-ticket-item.html
timovanasten: Jup, seems like a better place to put that variable. Will do.

PR #663, opened 2019-01-25
Fix validation of hidden fields

timovanasten:
Fix bug that prevented users from changing their profile if they registered without buying a ticket. Polymer had hidden the fields instead of removing them from the DOM, which caused them to still be validated by the .validate() method of iron-form. Adding the restamp attribute fixes this problem.

### DIFF
--- a/src/lancie-my-area/lancie-profile-edit.html
+++ b/src/lancie-my-area/lancie-profile-edit.html
@@ -85,7 +85,7 @@
           <paper-input label="Phone Number" name="phoneNumber" type="text" value="{{profile.phoneNumber}}" required></paper-input>
         </div>
         <div>
-          <template is="dom-if" if="[[addressNeeded]]">
+          <template is="dom-if" if="[[addressNeeded]]" restamp="true">
             <iron-icon icon="lancie:location-on"></iron-icon>
             <div class="group" >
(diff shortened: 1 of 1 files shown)


src/lancie-my-area/lancie-profile-edit.html
```diff
@@ -85,7 +85,7 @@
           <paper-input label="Phone Number" name="phoneNumber" type="text" value="{{profile.phoneNumber}}" required></paper-input>
         </div>
         <div>
-          <template is="dom-if" if="[[addressNeeded]]">
+          <template is="dom-if" if="[[addressNeeded]]" restamp="true">
```
martijnjanssen: I think we can even remove the dom-if here and replace it by hiding the div and setting the input to disabled if the address is not needed.

timovanasten: We could do that, but that comes down to the same thing in the end. This is just a bug fix for the current implementation.

martijnjanssen: dom-if's are still resource heavy, and this restamp statement is kind of a hack for the problem at hand.

Leaving this here can cause unclarity about implementation for the next committees.

PR #668, opened 2019-02-14
Update tournament sponsors and prizes

timovanasten:


timhuisman1:
I think you have linked to the wrong path in your code your images are in the wrong folder.

timovanasten:
Try running `yarn build optimize-images` first. Should fix it.

PR #670, opened 2019-02-26
Add timetable to the site

timovanasten:
vrijdag
zaterdag
zondag

### DIFF
--- a/src/lancie-home-page/lancie-home-page.html
+++ b/src/lancie-home-page/lancie-home-page.html
@@ -83,16 +83,20 @@
       <lancie-sponsors></lancie-sponsors>
     </lancie-section> -->
 
-    <lancie-section type="primary" header="Updates">
-      <div>
-        <p>
-          Don't worry, we hate junkmail too. We will not share your details with anyone.<br/>
-          You will only receive emails relevant to the event, such as when
-          the ticket sales are opened.
-        </p>
-        <lancie-subscribe-mail-form></lancie-subscribe-mail-form>
-      </div>
+    <lancie-section type="primary" header="Timetable">
+      <lancie-timeline></lancie-timeline>
     </lancie-section>
+
+    <!--<lancie-section type="primary" header="Updates">-->
+      <!--<div>-->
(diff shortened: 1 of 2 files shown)


src/lancie-home-page/lancie-home-page.html
```diff
@@ -83,17 +83,21 @@
...
-        <lancie-subscribe-mail-form></lancie-subscribe-mail-form>
-      </div>
+    <lancie-section type="primary" header="Timetable">
+      <lancie-timeline></lancie-timeline>
     </lancie-section>
 
+    <!--<lancie-section type="primary" header="Updates">-->
+      <!--<div>-->
+        <!--<p>-->
+          <!--Don't worry, we hate junkmail too. We will not share your details with anyone.<br/>-->
```
TimvdLippe: Don't think we can remove this?

martijnjanssen: I also prefer not to, but since we are adding an element, all colours have to be flipped around. I prefer not removing the email signup, but yeah, leaving it in is a ton of work.

We can undo it after the event.

TimvdLippe: Then we should refactor to use `:nth-child` instead.

timhuisman1: The subscribe to updates section is not really necessary at the moment right? Especially with the event being right around the corner and the fact that we won't be sending any more updates via this mailing list. If I recall correctly this section was commented out after the previous event as well.

TimvdLippe: It is your call, but I think it is a bad move to "simply the thing that bothers me right now and I will fix it later". It's like commenting out broken tests instead of fixing them.

But if you insist, please go ahead and make this change.

timovanasten: I did not remove the update form because it was in the way, but because I thought it had no value for the participants to be on the site when the event is happening. The timetable is probably the information the participants need most. So I guess that is more the discussion? Should the update form still be there?

PR #672, opened 2019-02-27
Update tournament rules

timovanasten:
Updated the rules with some input from experts.


timovanasten:
Offline review by Chris

PR #682, opened 2020-02-24
Feature schedule

bramvankooten:
Enable the tournament schedule and add the final tournament to the homepage


Sebastien1999:
Time schedule looks good! The "updates" widget should be removed, as we do not need it anymore and it clutters the screen.

PR #694, opened 2021-01-02
Add button to set captain of a team

DevDonut:


src/lancie-my-area/my-area-teams/team-member-item.html
```diff
@@ -70,6 +82,25 @@ <h2>Remove a team member</h2>
...
+      this.$.changeCaptainDialog.open();
+    },
+    submitChangeCaptain: function() {
+      this.$.ajaxChangeCaptain.generateRequest();
+    },
+    onChangeCaptain: function(event, request) {
+      if (request.succeeded) {
+        this.$.changeCaptainDialog.close();
+        this.fire('toast', {text: 'User successfully set as captain!'});
+        this.fire('remove-member');
```
DevDonut, line 95: I had a question about this line; I noticed this refreshes the member list, which is exactly what i needed, but i couldn't find the place where remove-member is defined. Does anyone know that?

skamoen: With this line you fire an event called "remove-member", which a parent element can catch an attach a listener. This happens here for the remove-member event: https://github.com/AreaFiftyLAN/lancie-frontend/blob/master/src/lancie-my-area/my-area-teams.html#L68

src/lancie-my-area/my-area-teams/team-member-item.html
```diff
@@ -70,6 +82,25 @@ <h2>Remove a team member</h2>
...
       }
     },
+    openChangeCaptainDialog: function() {
+      this.$.changeCaptainDialog.open();
+    },
+    submitChangeCaptain: function() {
+      this.$.ajaxChangeCaptain.generateRequest();
+    },
+    onChangeCaptain: function(event, request) {
+      if (request.succeeded) {
```
dsluijk, line 92: Not a huge issue, but this could be written in a more readable way.

- This has less branching, so it's easier to read.
- More importantly, the flow is directed at the happy path making it more linearly readable.

```
if (!request.succeeded) {
  const msg = request.request.status === 403 ? request.request.message : "Something went wrong setting this user as captain. Please try again.";
  this.fire('toast', {text: msg});
  return;
}
this.$.changeCaptainDialog.close();
this.fire('toast', {text: 'User successfully set as captain!'});
this.fire('remove-member');
```

If you don't like the ternary operator you can also just do this. Same effect, just a little more verbose.

```
if (!request.succeeded) {
  if (request.request.status === 403) {
    this.fire('toast', {text: request.request.message});
  } else {
    this.fire('toast', {text: "Something went wrong setting this user as captain. Please try again."});
  }
  return;
}
this.$.changeCaptainDialog.close();
this.fire('toast', {text: 'User successfully set as captain!'});
this.fire('remove-member');
```

dsluijk (approved):

ArdyZ:
@DevDonut and @dsluijk what should happen with this pull request? Can it be merged or should things change first?

PR #700, opened 2021-12-03
Create docker.yaml github workflow

julian9499:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10 AS builder
+FROM node:12 AS builder
```
martijnjanssen, line 1: Why only bump this version and not the second `FROM`?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,34 @@
+name: Docker
+on:
+  push:
+  schedule:
+  - cron: '54 2 2 * *'
```
martijnjanssen, line 5: Every 2nd day of the month on 02:54? Am I reading this right?

Dockerfile
julian9499: I only did this to manage build compatibility, but you are right I would suggest we update to node 12 across all the builders

julian9499: dockefiles*

.github/workflows/docker.yaml
julian9499: Leftover from another repo mb

bramvankooten: Guess Mark thought of that one 😆

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          images: ghcr.io/areafiftylan/lancie-frontend
+          tags: type=sha, prefix={{date 'YYYYMMDD'}}-
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          push: ${{ github.ref == 'refs/heads/master' }}
```
martijnjanssen, line 34: Push should only be on the `live` branch for this repo.

bramvankooten (approved):

PR #720, opened 2022-10-10
Update to the site

blibliboe:
Updating the site so it is correct for the 9th LANcie


src/lancie-home-page/lancie-home-page.html
```diff
@@ -135,7 +135,7 @@
           everything named above. AreaFiftyLAN will be an
           unforgettable event you do not want to miss! Are you ready for the
           launch? -->
-          <br /><sub>* CH members/alumni will get a BIG discount</sub>
```
ajaggoe, line 138: Why no big discount :(

ajaggoe (approved):
Ziet er goed uit, we moeten nog wel namen in de backend veranderen

PR #723, opened 2023-01-11
Fix routing to socials images

ajaggoe:
Fixed a bug where the links to the socials routed to the wrong images folder, this is now fixed (hopefully).


blibliboe (approved):
lgtm

PR #727, opened 2023-04-13
Hotfix sponsor

ajaggoe:
- Removed accidental sponsor as tournament sponsor
- Update other tournament sponsors with prizes
- Update Terms and Conditions


blibliboe (approved):
lgtm

